PR #113, opened 2021-12-17
New eff dev

gilbo:
I want to get this into master so that @yamaguchi1024 doesn't have to worry about the bug that was manifesting in `test_x86`.

The changes ought to be reasonably tightly grouped into commits, so it may be easiest to review changes commit-by-commit.


alexreinking (approved):

yamaguchi1024:
Cool, thanks!

PR #120, opened 2021-12-29
first cut at implementing winograd

yamaguchi1024:


requirements.txt
```diff
@@ -12,5 +12,6 @@ sympy==1.9
 tox==3.24.4
 yapf==0.31.0
 z3-solver==4.8.12.0
+torch
```
alexreinking, line 15: Need to set a version here to ensure consistency between developers and reproducibility generally.

alexreinking (changes requested):

alexreinking (approved):

PR #123, opened 2022-01-02
expand partial_eval to take boolean

yamaguchi1024:


SYS_ATL/LoopIR_scheduling.py
```diff
@@ -553,12 +553,12 @@ def __init__(self, proc, arg_vals):
 
         # Validate env:
         for k, v in self.env.items():
-            if not arg_types[k].is_indexable():
+            if not arg_types[k].is_indexable() and not arg_types[k].is_bool():
                 raise SchedulingError("cannot partially evaluate "
-                                      "numeric (non-index) arguments")
-            if not isinstance(v, int):
+                                      "numeric (non-index, non-bool) arguments")
+            if not isinstance(v, int) and not isinstance(v, bool):
```
alexreinking, line 559: Bool is a subtype of int in Python. This should just be `not isinstance(v, int)` and a comment should be left about intentionally excluding bool, too.

alexreinking (changes requested):
One bit of code cleanup needed, but otherwise LGTM

SYS_ATL/LoopIR_scheduling.py
yamaguchi1024: I see, that makes sense

alexreinking (approved):

alexreinking:
LGTM - go ahead and merge when the checks pass

yamaguchi1024:
> LGTM - go ahead and merge when the checks pass

Thanks!

PR #125, opened 2022-01-02
Handle unused control arguments(index, bool) in replace

yamaguchi1024:
Replace should unify the subprocedure with unused control arguments.
Currently, it does not handle unused buffer arguments (what does it even
    mean when the buffer argument was tensor or window) ?

### DIFF
--- a/SYS_ATL/LoopIR_unification.py
+++ b/SYS_ATL/LoopIR_unification.py
@@ -701,10 +701,13 @@ class Unification:
                                      stmt_block[0].srcinfo)
             elif fa.type == T.bool:
                 if fa.name in self.bool_holes:
-                    return self.bool_holes[fa.name]
+                    if self.bool_holes[fa.name] is False:
+                        return LoopIR.Const(False, T.bool, stmt_block[0].srcinfo)
+                    else:
+                        return self.bool_holes[fa.name]
                 else:
-                    # doesn't matter; argument un-used
-                    return LoopIR.Const(True, T.bool, stmt_block[0].srcinfo)
+                    # subprocedure argument must be consistent
+                    assert False, "bad case"
             elif fa.type == T.stride:
                 if fa.name in self.stride_holes:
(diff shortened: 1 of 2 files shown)


alexreinking:
Don't bother force pushing, because we're going to squash the commits anyway. Having the history around is helpful for review.

SYS_ATL/LoopIR_unification.py
```diff
@@ -701,10 +701,13 @@ def get_arg(fa):
                                      stmt_block[0].srcinfo)
             elif fa.type == T.bool:
                 if fa.name in self.bool_holes:
-                    return self.bool_holes[fa.name]
```
yamaguchi1024, line 704: @gilbo Can you take a look at this line? If the `self.bool_holes[fa.name]` was unchanged (False), that means it was an unused argument. Return `LoopIR.expr` in that case.

SYS_ATL/LoopIR_unification.py
```diff
@@ -701,10 +701,13 @@ def get_arg(fa):
...
+                    if self.bool_holes[fa.name] is False:
+                        return LoopIR.Const(False, T.bool, stmt_block[0].srcinfo)
+                    else:
+                        return self.bool_holes[fa.name]
                 else:
-                    # doesn't matter; argument un-used
-                    return LoopIR.Const(True, T.bool, stmt_block[0].srcinfo)
+                    # subprocedure argument must be consistent
+                    assert False, "bad case"
```
yamaguchi1024, line 710: @gilbo I'm not sure what was the original intention of this case? `self.bool_holes` is constructed by `subproc.args`, and this function `get_arg` is called with `subproc.args`, so there should be no case where `fa.name` is not in `self.bool_holes`.

gilbo:
This all looks fine.  The code before this change was definitely buggy/inconsistent.

gilbo (approved):

PR #130, opened 2022-01-11
Fix parse_fragment and expand_dim to take variables

yamaguchi1024:


yamaguchi1024:
The CI failure is from gemmini tests as expand_dim now uses old EffectCheck for the bounds check. Can we merge this as is or disable gemmini tests (which sounds bad?)

SYS_ATL/API.py
```diff
@@ -519,10 +519,13 @@ def expand_dim(self, stmt_pat, alloc_dim_pat, indexing_pat):
...
             s = self._find_stmt(stmt_pat, body=loopir.body, default_match_no=None)[i]
-            alloc_dim = parse_fragment(loopir, alloc_dim_pat, s)
-            indexing  = parse_fragment(loopir, indexing_pat, s)
+            alloc_dim = parse_fragment(loopir, alloc_dim_pat, None)
+            indexing  = parse_fragment(loopir, indexing_pat, None)
             loopir = Schedules.DoExpandDim(loopir, s, alloc_dim, indexing).result()
 
+        # Running checkeffect here is necessary for bounds checking
+        CheckEffects(loopir)
+
```
alexreinking, line 528: Is this the line that's causing the Gemmini failures?

alexreinking: It appears to be.

yamaguchi1024: Yeah, we need this line for the correctness of `expand_dim`

alexreinking:
This is how we ought to handle this...

1. We will add an `unsafe_disable_checks` parameter to `expand_dim()`. It is `False` by default.
2. When `unsafe_disable_checks` is `True`, then we skip the call to `CheckEffects`.
3. We will open an issue to update the Gemmini tests.
4. The Gemmini tests will be updated to set `unsafe_disable_checks` to `True`. Every such instance will be marked by a comment that reads: `# FIXME (#X): these schedules need to be updated to work with the new effect checking system`, where `X` is the issue number created in step 3.

yamaguchi1024:
I see, that totally makes sense

tests/gemmini/conv/test_gemmini_conv_no_pad.py
```diff
@@ -188,7 +188,7 @@ def test_conv_3():
     conv = conv.fission_after('for ocol_o in _:_ #0')
     conv = conv.reorder('orow', 'ocol_o')
     conv = conv.split('orow', 28, ['orow_o', 'orow_i'], perfect=True)
-    conv = conv.expand_dim('i_s: i8[_]', '30', 'krow + orow_i')
+    conv = conv.expand_dim('i_s: i8[_]', '30', 'krow + orow_i', unsafe_disable_checks=True)
```
alexreinking, line 191: Where are the `FIXME` comments? Has an issue been opened?

alexreinking (changes requested):

SYS_ATL/API.py
```diff
@@ -519,10 +520,14 @@ def expand_dim(self, stmt_pat, alloc_dim_pat, indexing_pat):
         loopir = self._loopir_proc
         for i in range(0, stmts_len):
             s = self._find_stmt(stmt_pat, body=loopir.body, default_match_no=None)[i]
-            alloc_dim = parse_fragment(loopir, alloc_dim_pat, s)
-            indexing  = parse_fragment(loopir, indexing_pat, s)
+            alloc_dim = parse_fragment(loopir, alloc_dim_pat, s, scope="before")
+            indexing  = parse_fragment(loopir, indexing_pat, s, scope="before_after")
             loopir = Schedules.DoExpandDim(loopir, s, alloc_dim, indexing).result()
 
+        if unsafe_disable_checks == False:
```
alexreinking, line 527: `if not unsafe_disable_checks:`

tests/gemmini/conv/test_gemmini_conv_no_pad.py
yamaguchi1024: Ah forgot that

alexreinking:
> * We will open an issue to update the Gemmini tests.
> 
> * The Gemmini tests will be updated to set `unsafe_disable_checks` to `True`. Every such instance will be marked by a comment that reads: `# FIXME (#X): these schedules need to be updated to work with the new effect checking system`, where `X` is the issue number created in step 3.

I have updated your PR with the issue number cross-reference... LGTM once CI passes.

yamaguchi1024:
Thanks!

alexreinking (approved):

PR #136, opened 2022-01-14
implement lift_alloc_simple

yamaguchi1024:


SYS_ATL/API.py
```diff
@@ -1031,6 +1031,22 @@ def par_to_seq(self, par_pattern):
...
 
+    def lift_alloc_simple(self, alloc_site_pattern, n_lifts=1):
+        if not is_pos_int(n_lifts):
+            raise TypeError("expected second argument 'n_lifts' to be "
+                            "a positive integer")
+
+        stmts_len = len(self._find_stmt(alloc_site_pattern, default_match_no=None))
+        loopir = self._loopir_proc
+        for i in range(0, stmts_len):
+            s = self._find_stmt(alloc_site_pattern, body=loopir.body, default_match_no=None)[i]
```
alexreinking, line 1042: I find this loop construction very confusing and I'm not 100% sure it's correct... at the very least there should be some test that lifts multiple allocations with this directive.

SYS_ATL/API.py
```diff
@@ -1031,6 +1031,22 @@ def par_to_seq(self, par_pattern):
 
         return Procedure(loopir, _provenance_eq_Procedure=self)
 
+    def lift_alloc_simple(self, alloc_site_pattern, n_lifts=1):
```
alexreinking, line 1034: I don't understand why this is a totally different API from `lift_alloc` rather than an option.

alexreinking (changes requested):
I'm interested in @gilbo's thoughts on this, too.

SYS_ATL/API.py
yamaguchi1024: I'm planning to eventually implement `lift_alloc` using `lift_alloc_simple`, `expand_dim`, and `rearrange_dim` . I think we talked about this in a meeting in December

alexreinking: Okay, we can discuss naming later, then. It seems confusing to me to have both `lift_alloc` and `lift_alloc_simple` in the same system.

yamaguchi1024: I think our plan is to eventually replace `lift_alloc` with `lift_alloc_simple`. I think it's fine for both to coexist for a while before we rewrite all the tests

yamaguchi1024:
@alexreinking I addressed your comments, thanks! `n_lifts==0` case was checked in both API.py interface and in assert in `__init__`.

SYS_ATL/API.py
alexreinking: Sounds good to me!

alexreinking (approved):
Looks great!

PR #137, opened 2022-01-14
Implement remove_loop

yamaguchi1024:


SYS_ATL/LoopIR_scheduling.py
```diff
@@ -1679,6 +1679,38 @@ def map_s(self, s):
...
+    def map_s(self, s):
+        if s is self.stmt:
+            # Check if we can remove the loop
+            # Conditions are:
+            # 1. Body does not depend on the loop iteration variable
+            # 2. Body is idemopotent
+            # 3. The loop runs at least once
+            # we should run something similar to old effectcheck's
+            # is_pos_int to check 3, but that's a TODO.
+            # for now, we'll check 1 and 2.
```
alexreinking, line 1699: Not including check (3) makes this unsafe, which is not great. You should _make_ it safe by wrapping the loop body in an `if` that checks the loop extent is positive at runtime. It's possible the simplifier could (in some cases) resolve the condition to `True` and eliminate the branch. It's better to have a safe scheduling language that could be improved than an unsafe scheduling language that goes faster when it happens to work, sometimes.

Finally, TODO comments should be formatted as `# TODO: ...` to make them easy to search for. After implementing the runtime check, I suggest:

```
# TODO: (3) could be checked statically using something similar to the legacy is_pos_int.
```

SYS_ATL/API.py
```diff
@@ -1070,6 +1070,19 @@ def double_fission(self, stmt_pat1, stmt_pat2, n_lifts=1):
 
         return Procedure(loopir, _provenance_eq_Procedure=self)
 
+    def remove_loop(self, loop_pattern):
```
alexreinking, line 1073: I think this name suggests an unsafe operation that outright _deletes_ some code. This is really a special case of loop unrolling where the upper bound doesn't matter (apart from being positive) and the body is checked for idempotence. I won't block the PR on this, but a major part of me wonders whether the user experience would be better to roll this into `unroll_loop`.

To see what I mean:

```
for i in seq(0, N): S  =(unroll)>  S0 S1 ... SN-1 (when N > 0)  =(drop idempotent)>  S0
```

SYS_ATL/API.py
```diff
@@ -1070,6 +1070,19 @@ def double_fission(self, stmt_pat1, stmt_pat2, n_lifts=1):
...
+        if not isinstance(loop_pattern, str):
+            raise TypeError("expected first arg to be a string")
+
+        stmts_len = len(self._find_stmt(loop_pattern, default_match_no=None))
+        loopir = self._loopir_proc
+        for i in range(0, stmts_len):
+            s = self._find_stmt(loop_pattern, body=loopir.body, default_match_no=None)[i]
```
alexreinking, line 1080: Again, there should be a test with multiple targets to exercise this.

alexreinking (changes requested):

SYS_ATL/LoopIR_scheduling.py
yamaguchi1024: Well, `remove_loop` is already happening as part of `fission_after` without (3). We have `is_pos_int` in the old effect check, but that interface is not exposed to LoopIR_scheduling.py, and I think @gilbo is working on making a similar interface with new effect check. I will need to implement whole SMT solver logic here if I want to check (3), which will be a huge duplication.

SYS_ATL/API.py
yamaguchi1024: I see, I think that makes sense and I don't have a strong opinion with either interface. `remove_loop` might be slightly more explicit than `unroll` just magically removing the loop for the user? Not sure

alexreinking: I don't see this as magic for `unroll`... there are three cases:

1. The user asked to unroll a loop of constant, positive, extent... do the obvious thing.
2. The user asked to unroll a loop of non-constant extent _and_ the loop body is idempotent... replace with `if extent > 0: s[i/extent]`
3. The user asked to unroll a loop that cannot be unrolled... `raise SchedulingError`.

SYS_ATL/LoopIR_scheduling.py
alexreinking: I'm saying to check (3) at _runtime_, which only involves the following rewrite:

```
for i in seq(0, N):  =>  if N > 0:
  idempotent body          idempotent body
```

SYS_ATL/API.py
yamaguchi1024: I see. Let's put this on the agenda on Monday? I'm fine with both, we can sort out `remove_loop` and merge it to `unroll` afterward if necessary.

alexreinking: Sure, sounds good to me.

SYS_ATL/LoopIR_scheduling.py
yamaguchi1024: Ah I see! That makes sense, thanks!

alexreinking (approved):

PR #140, opened 2022-01-16
Implement fission_after_simple

yamaguchi1024:
After merging this and rearrange_dim, I am planning to implement:
1. change `lift_alloc` to be a composition of `lift_alloc_simple`, `expand_dim`, and `rearrange_dim`
2. change `fission_after` to be a composition of `fission_after_simple` and `remove_loop`

### DIFF
--- a/SYS_ATL/LoopIR_scheduling.py
+++ b/SYS_ATL/LoopIR_scheduling.py
@@ -1847,7 +1847,7 @@ class _DoFissionAfterSimple:
     # see map_stmts comment
     def map_s(self, s):
         if s is self.tgt_stmt:
-            #assert self.hit_fission == False
+            assert self.hit_fission == False
             self.hit_fission = True
             # none-the-less make sure we return this statement in
             # the pre-fission position
@@ -1857,8 +1857,7 @@ class _DoFissionAfterSimple:
 
             # first, check if we need to split the body
             pre, post = self.map_stmts(s.body)
-            fission_body = (len(pre) > 0 and len(post) > 0 and self.n_lifts > 0)
-            if fission_body:
+            if pre and post and self.n_lifts > 0:
                 self.n_lifts -= 1
                 self.alloc_check(pre, post)
                 pre = LoopIR.If(s.cond, pre, [], None, s.srcinfo)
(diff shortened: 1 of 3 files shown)


yamaguchi1024:
Why does the build not being triggered??

alexreinking:
I'm not sure... maybe the workflow isn't running on merge commits for some reason?

alexreinking:
Closing and reopening the PR seemed to restart GHA... I'll Google around a bit to find a permanent fix

SYS_ATL/LoopIR_scheduling.py
```diff
@@ -1799,6 +1799,127 @@ def map_s(self, s):
...
+            # then we need to gather together the pre and post.
+            single_stmt = LoopIR.If(s.cond, body, orelse, None, s.srcinfo)
+
+        elif isinstance(s, LoopIR.ForAll) or isinstance(s, LoopIR.Seq):
+
+            # check if we need to split the loop
+            pre, post = self.map_stmts(s.body)
+            do_fission = (len(pre) > 0 and len(post) > 0 and
+                          self.n_lifts > 0)
+            if do_fission:
```
alexreinking, line 1894: This and other checks like it can be replaced with `if pre and post and self.n_lifts > 0:`.

In general, don't use `len()` to check if a list is empty or not.

SYS_ATL/LoopIR_scheduling.py
```diff
@@ -1799,6 +1799,127 @@ def map_s(self, s):
...
+            pre, post       = self.map_s(orig_s)
+            pre_stmts      += pre
+            post_stmts     += post
+
+        return (pre_stmts, post_stmts)
+
+    # see map_stmts comment
+    def map_s(self, s):
+        if s is self.tgt_stmt:
+            #assert self.hit_fission == False
```
alexreinking, line 1850: Why would this assert fail?

alexreinking (changes requested):
Just fix the list emptiness checks and we're good to go

SYS_ATL/LoopIR_scheduling.py
yamaguchi1024: Because this pass should be executed just once where object s exactly matches with self.tgt_stmt.

alexreinking (approved):
LGTM

PR #141, opened 2022-01-18
AMX Matmul Scheduling

skeqiqevian:


SYS_ATL/LoopIR_compiler.py
```diff
@@ -676,6 +676,7 @@ def comp_s(self, s):
                         assert isinstance(s.args[i], LoopIR.WindowExpr)
                         data, _ = self.window_struct_fields(s.args[i])
                         d[f'{arg_name}_data'] = data
+                        d[f'{arg_name}_int'] = self.env[s.args[i].name]
```
yamaguchi1024, line 679: Note that we needed this because AMX tile require `const`

yamaguchi1024 (approved):
LGTM, thanks!

SYS_ATL/memory.py
```diff
@@ -120,7 +120,7 @@ def alloc(cls, new_name, prim_type, shape, srcinfo):
         if len(shape) == 0:
             return f"{prim_type} {new_name};"
 
-        return (f"{prim_type} *{new_name} = "
+        return (f"{prim_type} *{new_name} = ({prim_type} *)"
```
alexreinking, line 123: This not needed... why was this added?

tests/amx/harness_amx.py
```diff
@@ -267,4 +269,4 @@ def stop_timer(self, name, msg):
 
     def add_body(self, lines):
         assert type(lines) is list
-        self.body += lines
+        self.body += lines
```
alexreinking, line 272: Why did you remove the newline at the end of the file?

tests/amx/harness_amx.py
```diff
@@ -13,8 +13,10 @@
     pytest.skip("skipping AMX tests; could not find sde",
                 allow_module_level=True)
 
-CC_BAREMETAL        = os.getenv('CLANG') or os.getenv('CC', 'clang-13')
+# CC_BAREMETAL        = os.getenv('CLANG') or os.getenv('CC', 'clang-13')
+CC_BAREMETAL = "clang++"
 CFLAGS_BAREMETAL    = ' '.join([
+                        f'-std=c++11',
```
alexreinking, line 19: No.

tests/amx/harness_amx.py
```diff
@@ -13,8 +13,10 @@
     pytest.skip("skipping AMX tests; could not find sde",
                 allow_module_level=True)
 
-CC_BAREMETAL        = os.getenv('CLANG') or os.getenv('CC', 'clang-13')
+# CC_BAREMETAL        = os.getenv('CLANG') or os.getenv('CC', 'clang-13')
+CC_BAREMETAL = "clang++"
```
alexreinking, line 17: No. We generate C code, not C++.

alexreinking:
This should not have been merged in this state. See my comments above. We specifically generate C code... adding `-std=c++11` to a C++ compiler in our tests is a sure-fire way to introduce bugs related to small differences in C / C++ semantics. It's one thing to compile test code with a C++ compiler and link it to a SYSTL-generated C library, but a very different thing to compile SYSTL-generated code as C++.

yamaguchi1024:
We needed to use C++ compiler for `constexpr`. I think using C++ compiler for AMX test specifically is fine

alexreinking:
As I said:

> It's one thing to compile **test code** with a C++ compiler and link it to a SYSTL-generated C library, but a very different thing to compile SYSTL-generated code as C++.

A change was made to LoopIR_compiler in service of making SYSTL code compilable with a C++ compiler. This not an acceptable change.

yamaguchi1024:
The change is not made to `LoopIR_compiler`?

yamaguchi1024:
It's in `harness_amx.py`

SYS_ATL/LoopIR_compiler.py
alexreinking: @yamaguchi1024 - This is LoopIR_compiler.py

SYS_ATL/memory.py
alexreinking: @yamaguchi1024 - this is injected into generated code that should not be compiled with C++.

SYS_ATL/LoopIR_compiler.py
alexreinking: You can't pass `const` between C++ and C. It's an ABI incompatibility.

SYS_ATL/memory.py
yamaguchi1024: Clang++ was more strict about the type and IIRC we needed that casting to compile with `clang++` so yeah this change is related to us needing to use C++ compiler, but I don't see a reason why explicit casting is a problem.

tests/amx/harness_amx.py
yamaguchi1024: Constexpr is supported from C++11

alexreinking: This is okay for the _test code_ only, _not_ for code that is generated by SYSTL. This will need to be re-written to use two different compilers.

SYS_ATL/LoopIR_compiler.py
yamaguchi1024: There is no external linkage here or anything, so I'm not sure what is the problem?

SYS_ATL/memory.py
alexreinking: It's because C and C++ are different languages and C++ is _not_ a superset of C. It's not that Clang++ is "more strict", rather Clang++ is following _different rules_. Compiling our generated code with a C++ compiler is explicitly not supported and ever doing so increases the risk of a language incompatibility bug that outright kills our portability argument.

SYS_ATL/LoopIR_compiler.py
alexreinking: There _has to be_ external linkage here, because we must not compile our generated code as C++. The headers have (or ought to have)
```
#ifdef __cplusplus
extern "C" {
#endif 

...

#ifdef __cplusplus
}
#endif
```
in them.

SYS_ATL/memory.py
yamaguchi1024: I agree that we shouldn't explicitly support C++. I can try with @skeqiqevian not to use C++ compiler (looking at the current code it seems like C `const` would work? not sure) so I think this change can be reverted, but I think the change to `LoopIR_compiler.py` need to stay as we need a way not to compile buffer as a pointer.

SYS_ATL/libs/memories.py
```diff
@@ -231,15 +231,20 @@ def window(cls, basetyp, baseptr, indices, strides, srcinfo):
...
+        global num_amx_tiles_alloced
+        num_amx_tiles_alloced += 1
+        return f"constexpr int {new_name} = {num_amx_tiles_alloced-1};"
 
     @classmethod
     def free(cls, new_name, prim_type, shape, srcinfo):
-        return ""
+        global num_amx_tiles_alloced
+        num_amx_tiles_alloced -= 1
+        return ""
```
alexreinking, line 250: Newlines at ends of files.

SYS_ATL/LoopIR_compiler.py
yamaguchi1024: Ok. If we use C `const` and keep this code, would that be an issue?

tests/amx/harness_amx.py
yamaguchi1024: This I'm not sure. @skeqiqevian ?

SYS_ATL/memory.py
alexreinking: > I think the change to LoopIR_compiler.py need to stay as we need a way not to compile buffer as a pointer.

I am not exactly clear what's happening here, either. Can you show me an example of the generated code? If this is equivalent to `(intptr){arg_data}`, then just use that...

yamaguchi1024:
@alexreinking Let me discuss this with @skeqiqevian and we'll get back to you. I think it's possible to just use the C compiler since we decided not to do constexpr magic here (using C++ compiler decision was made when we thought that we needed to do fancy stuff with metaprogramming, but it's not the case anymore). This discussion was super helpful, thank you!

SYS_ATL/LoopIR_compiler.py
alexreinking: Depends where the const is... an extern-C const function parameter has the same ABI as a non-const parameter in both languages. However, changing a _struct field_ to `const` has a different ABI.

PR #142, opened 2022-01-21
AMX Matmul Scheduling, but with C++ changes removed

skeqiqevian:


yamaguchi1024:
Seems like you need to pull from master?

skeqiqevian:
@yamaguchi1024 ack sorry about that. I merged master in now.

yamaguchi1024:
@skeqiqevian Also, could you remove the unrelated code as much as possible? (e.g, changes to `tests/amx/amx_experiment.c ` seems not related to C++->C change?)

yamaguchi1024:
> @skeqiqevian Also, could you remove the unrelated code as much as possible? (e.g, changes to `tests/amx/amx_experiment.c ` seems not related to C++->C change?)

NVM

yamaguchi1024:
@alexreinking 
This PR removes all the concerns you had about the other PR, except the change to `LoopIR_compiler.py` where we introduced `_int` thing ( https://github.com/ChezJrk/SYS_ATL/blob/master/SYS_ATL/LoopIR_compiler.py#L679 ).

Let me try to explain why we still need that change even though we are using clang now. Btw, we need to use `clang-13` because that is the only compiler that supports the AMX instructions currently.

AMX instructions such as `_tile_zero` take only integer constants. So the following code works in C++, but not in C, because `const` is not a compile-time constant in C.
```
const int tile = 2;
_tile_zero(tile);
```

Thus, we decided to use C macro, so that it becomes
```
#define tile 2
_tile_zero(tile);
#undef tile
```

We want to treat AMX `tile` as a buffer like everything else, but we don't want to index into it (using `.data` or `_data` would give `tile[0]` which doesn't make any sense here and it's certainly not a constant). That's why we need to have a codegen path that compiles a buffer without indexing.

alexreinking (approved):
This looks good... thanks!

PR #144, opened 2022-01-24
Bug in InlineWindow()

gilbo:
I wrote a test to exercise the problem with `InlineWindow()`.  Make sure that all possible occurrences of the base window/buffer get substituted, not just expressions.  I recommend auditing `LoopIR` to be sure they're all covered.

### DIFF
--- a/SYS_ATL/LoopIR_scheduling.py
+++ b/SYS_ATL/LoopIR_scheduling.py
@@ -756,9 +756,33 @@ class _InlineWindow(LoopIR_Rewrite):
         # repair effects...
         self.proc = InferEffects(self.proc).result()
 
+
+    def calc_idx(self, idxs):
+        assert (len([w for w in self.win_stmt.rhs.idx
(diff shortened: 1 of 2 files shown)


yamaguchi1024:
Oh, thanks!

yamaguchi1024 (approved):
Fixed the bug, thanks for catching!

SYS_ATL/LoopIR_scheduling.py
```diff
@@ -756,10 +756,34 @@ def __init__(self, proc, stmt):
...
+            else:
+                new_idxs.append(w.pt)
+
+        return new_idxs
+
     def map_s(self, s):
         if s is self.win_stmt:
             return []
 
+        if isinstance(s, LoopIR.Assign) or isinstance(s, LoopIR.Reduce):
```
alexreinking, line 780: This should have been `isinstance(s, (LoopIR.Assign, LoopIR.Reduce))`

PR #146, opened 2022-01-27
Added schedule frontend sketch

yamaguchi1024:
This PR is for review and discussion


examples/schedule_sketch.py
```diff
@@ -1,4 +1,90 @@
...
+        else:
+            proc = proc.simple_fission()
+
+# -----------------------------------------------------------------------------------------
+# Proposal 2 is to return the tuple of added statement and proc from scheduling directives.
+# Example code:
+def fission(proc, stmt, n_lifts):
+    # I guess this is more explicit, but can confuse users
+    for i in range(0, n_lifts):
+        (proc, stmt) = proc.simple_fission(stmt)
```
yamaguchi1024, line 52:
```
for j:
  for i in seq(0, n):
    s1 <--- stmt
    s2
----------->
for j:
  for i in seq(0, n): <---- stmt
    s1
  for i in seq(0, n):
    s2
----------->
for j: <---- stmt
  for i in seq(0, n):
    s1
for j:
  for i in seq(0, n):
    s2
```

examples/schedule_sketch.py
```diff
@@ -1,4 +1,90 @@
...
+
+# -----------------------------------------------------------------------------------------
+# Proposal 1 is to record the previously added statement in proc.
+# Example code:
+@proc
+def foo():
+    ...
+foo = foo.simple_fission('a = b[_]') # We don't need to modify the existing interface
+
+def fission(proc, stmt, n_lifts): # Users can write user-defined schedule in Python
```
yamaguchi1024, line 36: `simple_fission(proc, stmt)`
`fission(proc, stmt, n_lifts=3`

^ We should keep the same interface between primitive operator and user-defined operator

yamaguchi1024:
```
def repeat(proc, stmt, sched, n):
  for i in range(0, n):
    (proc, stmt) = sched(proc, stmt)
```
and you can call `repeat` as
`foo = repeat(proc, stmt, sched, 6)`

alexreinking: It should really be something like this:
```
def repeat(sched, n, *args):
  for i in range(n):
    args = sched(*args)
  return n
```

alexreinking: The point being that the interface for any particular scheduling directive isn't really relevant to the idea of repeating it.

yamaguchi1024: totally agreed

examples/schedule_sketch.py
```diff
@@ -1,4 +1,90 @@
...
+foo = foo.schedule(fission, ['for i in _:_', 3])
+
+# -----------------------------------------------------------------------------------------
+# The advantage of Proposal 2 and 3 are that the pointer (stmt and g) is explicit to the user,
+# so they can implement full fission with remove_loop like this. This is using Proposal 2 syntax.
+def fission(proc, stmt, n_lifts):
+    for i in range(0, n_lifts):
+        (proc, stmt) = proc.simple_fission(stmt)
+        (proc, stmt) = proc.remove_loop(stmt) # This removes the "pre" loop
+        (proc, stmt) = proc.remove_loop(stmt+1) # This should remove the "post" loop
```
yamaguchi1024, line 76: Comment from Gilbert: this should be `next()` or `next(5)` instead of `+1`, because `+1` is ambiguous.

examples/schedule_sketch.py
yamaguchi1024: Comment from Alex: we should think about how to handle "outdated" pointers. Should they be invalid after the first use? Should "pointers" hold data structure other than a mere object (such as traversal path) to make it re-useable?

examples/schedule_sketch.py
yamaguchi1024: Also comment from Gilbert: I should add more examples other than fission

examples/schedule_sketch.py
```diff
@@ -1,4 +1,576 @@
...
+    j_loop_var = stmt.body()[0].iter
+    proc, i_loop_out = proc.split(i_loop_var, iter_sizes[0], new_names[0], tail=tail)
+    proc, j_loop_out = proc.split(j_loop_var, iter_sizes[1], new_names[1], tail=tail)
+    i_loop_in = i_loop_out.body()[0]
+    proc, _ = proc.reorder(i_loop_in.iter, j_loop_out.iter)
+
+    return proc
+
+
+# Lift this statement as up as possible (is this English?)
```
alexreinking, line 43: > is this English?

I would say "Lift this statement as _high_ as possible".

yamaguchi1024:
Closing as this is no longer relevant

PR #148, opened 2022-01-28
Rework testing to use common infrastructure 

alexreinking:
This PR implements several Pytest fixtures that make it easier to write and manage tests. Once CI passes, I'll start writing real documentation here.

The gist is that you should now add any of the following arguments to your tests, as needed:

1. `golden`
    1.  If you want to compare some string output, this will automagically be the string associated with the current test.
    2. Pass `--update-golden` to Pytest to replace all the golden outputs.
2. `compiler`
    1. If you need to compile and either **load a DLL** or **run a test executable**, this will handle it.
3. `sde64`
    1. If you want to run something the `compiler` returned in the SDE emulator, this is how you do it.

You can now also mark a test as requiring a specific ISA to function. Use `@pytest.mark.isa('AVX512f')` to indicate that the test requires AVX512. The full list of features (along with their detection status) can be printed like so:

```
python -c 'import cpufeature, pprint; pprint.pprint(cpufeature.CPUFeature)'
```

----

This PR uncovered and fixed these bugs:

1. The UAST pretty printer wasn't using its ChainMap environments correctly. We had a similar bug in the LoopIR pretty printer a while ago, but it wasn't fixed in UAST.
2. The `extract_method` scheduling directive produced new methods with undefined argument ordering. It now sorts the arguments by name.
3. Coverage data was being collected about `@proc`, `@instr`, and `@config` decorated functions and classes.

---

This PR also enables AMX testing via SDE on GitHub Actions. We should be able to re-enable macOS testing in a follow-up PR.

Still have to rework the gemmini tests, since I don't have chipyard set up anymore and they were never run on CI.


yamaguchi1024:
I am a bit confused. Most of the test changes seem to be just formatting change, and they don't seem to be using the new `golden` or `compiler` infrastructure?

alexreinking:
> I am a bit confused. Most of the test changes seem to be just formatting change, and they don't seem to be using the new `golden` or `compiler` infrastructure?

I don't know what to say, but to look again? Every single test outside of `tests/gemmini` was converted.

alexreinking:
It might be useful to walk through the commits and look at the ones that explicitly add golden outputs for tests. I let my editor run its formatting before making functional changes to files so they could be evaluated separately.

alexreinking:
> Most of the test changes seem to be just formatting change

Also, this is another major reason to deal with #59 since diffs between two files that have been autoformatted with the same tool and settings include only functional changes.

yamaguchi1024:
Could you please not run your editor's autoformatting, especially when it contains a functional change to many files? I recall @gilbo was not in favor of autoformatting the code (from the discussion in November), and IIRC, that discussion of whether we should use any autoformatting was still unfinished. I'm personally totally fine with running the autoformatting to all the files in the separate PR, but the issue of code formatting is not related to the main change in this PR, I think.

alexreinking:
It's way too late for that, unfortunately. It would cost me several hours of work to undo that at this point.

yamaguchi1024:
```
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['ctest', '-C', 'Release', '--build-and-test', '/tmp/pytest-of-yuka/pytest-2/test_winograd0', '/tmp/pytest-of-yuka/pytest-2/test_winograd0/build', '--build-generator', 'Ninja', '--build-options']' returned non-zero exit status 1.

/usr/lib/python3.9/subprocess.py:528: CalledProcessError
```
^ This is the error I got from my environment

alexreinking:
@yamaguchi1024 - can you run:
```
$ cmake --version
$ ctest --version
```
If it's below 3.21, you will need to upgrade.

yamaguchi1024:
Yeah it is 3.16, will upgrade later

yamaguchi1024:
@alexreinking upgraded them and both cmake and ctest versions are 3.22.2, but they still give the same error. Do you know what could be an other issue?

alexreinking:
Can you try manually running the failing command and posting the full output? e.g.

```
ctest -C Release --build-and-test /tmp/pytest-of-yuka/pytest-2/test_winograd0 /tmp/pytest-of-yuka/pytest-2/test_winograd0/build --build-generator Ninja --build-options
```

yamaguchi1024:
Ok I just had to install ninja. Does README or requirements.txt document these requirements?

alexreinking:
You can technically override it by running `export CMAKE_GENERATOR="Unix Makefiles"` but you're right that this should all be documented.

alexreinking:
@yamaguchi1024 - I have added documentation to the README. Since all it took was installing the new dependencies and because it only affects the test suite, let's get this merged soon. I would like to start writing new tests (e.g. for the `lift_if` bug I found) using these helpers ASAP.

yamaguchi1024:
@alexreinking I'm good with these changes, but haven't we talked on Friday that we should ping users and check that it works on their machine before merging?

yamaguchi1024:
Also pinging @gilbo , you said that you wanted to go through the changes but is that done?

alexreinking:
I appreciate the reasoning for ringing the alarm bells, but I don't actually think it will be very useful to delay the PR. That's because there are a number of mitigating factors...

1. This does not introduce _any_ breaking changes to the main SYSTL library, so no one's code needs to be updated.
2. This will also not break _any_ additional tests that users have added. They will see failures in the other tests, but it will not block their (private) progress.
    1. They shouldn't be developing inside of `tests/` anyway. They should be building and installing our package into a virtual environment and writing code like what's in `apps/`.
3. The steps to get to a sane build environment on CI are completely standard for any mixed Python and C project. CI also shows things are working at least as well as they have been on both Linux and macOS.
4. This is a big change and we don't want it to go stale.

yamaguchi1024 (approved):

PR #152, opened 2022-02-02
Fix lift if

alexreinking:
I rewrote `lift_if` to support lifting an if statement past another if statement, which would previously _delete_ parent ifs. I have written a handful of tests (and will write more before merging). Opening this to get CI going / solicit more test cases in reviews.

**Update:** tests cover 100% of my implementation. This is obviously not a guarantee of correctness, but it's clearly better than it was.


SYS_ATL/new_eff.py
```diff
@@ -1156,8 +1156,8 @@ class ES(Enum):
 class SchedulingError(Exception):
     def __init__(self, message, **kwargs):
         ops = self._get_scheduling_ops()
-        # TODO: include outer ops in message
```
yamaguchi1024, line 1159: LGTM if this change in new_eff.py don't cause merge conflict with @gilbo 's branch

alexreinking: I can revert that for now. It was just to get a better error message while I was debugging.

yamaguchi1024 (approved):
LGTM, thanks!

PR #153, opened 2022-02-02
Remove dependency on cpufeature package.

alexreinking:
Replaces the `cpufeature` package with platform-specific logic to detect ISAs.

* If the `SYSTL_OVERRIDE_CPUINFO` environment variable is set to a space-separated list, uses its value.
* On Linux, parses `/proc/cpuinfo`.
* On macOS, queries x86-specific `sysctl` entries.

In all other cases, conservatively assumes that ISAs are _not_ available.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,6 @@ asdl==0.1.5
 astor==0.8.1
 build==0.7.0
 coverage==6.3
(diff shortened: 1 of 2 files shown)


gilbo (approved):
Seems fine; not scrutinizing the command line calls

PR #158, opened 2022-02-15
Force C/C++ code to be formatted by GitHub Actions

alexreinking:
Trying to dip our toes into automatic formatting. This new workflow will automatically run `clang-format` on all our C/C++ files and commit the result to the PR. Later, we can add Python autoformatting to this workflow.


yamaguchi1024:
Also, if CI does not trigger CI when linting the code, do we always need to trigger CI by pushing empty commit after formatting?

alexreinking:
> Does clang-format apply to new C files as well?

Yes, it applies to the whole repository except `dependencies/` and `tests/gemmini/gemmini-rocc-tests/` (i.e. third party code modulo small patches).

> Also, if CI does not trigger CI when linting the code, do we always need to trigger CI by pushing empty commit after formatting?

Yes, but that's not a problem. The alternative is to error when the code is improperly formatted, but then you have to do the work of running clang-format and pushing that, anyway. If the code is properly formatted to begin with, no new commits will be created. So no matter what, if the code is improperly formatted, then you will need to push _something_ (either an empty commit or a fix commit). This way is just less work (`git commit --allow-empty -m "trigger gha"`).

yamaguchi1024:
Is it not possible to trigger CI automatically after formatting? (no worries if that's difficult)

alexreinking:
> Is it not possible to trigger CI automatically after formatting? (no worries if that's difficult)

It is potentially dangerous if a bug in clang-format is non-idempotent on a source file.

yamaguchi1024 (approved):

PR #164, opened 2022-02-23
remove scipy and torch from requirements

yamaguchi1024:


alexreinking:
You're missing `import pytest`.

alexreinking:
Is `test_winograd.py` worth keeping at all? It looks from the comments like it wasn't even a correct test to begin with. It would be nicer to fix the test (perhaps by validating the values offline and then committing it as the golden output), but we could also just delete it.

yamaguchi1024:
I would like to keep the test. It's at least compliant with what Tian had implemented for his job.

yamaguchi1024:
Although I do agree that it can be optimized more.

yamaguchi1024:
And, it is not that "it was not the correct test, to begin with". It's just that we can expand the matrix thus can delete zero coefficients so that it becomes more optimal.

alexreinking:
What does `# TODO: finish test, validate output (returns NaN right now)` mean, then?

yamaguchi1024:
The test needs to be validated, but removing the torch dependency is a different issue. Python 3.10 doesn't support torch yet (and torch is a huge dependency), and not all the users need to run winograd test.

alexreinking:
> The test needs to be validated, but removing the torch dependency is a different issue. Python 3.10 doesn't support torch yet (and torch is a huge dependency)

It's not that different. This way is somewhat vacuous. We could also "remove" all of our dependencies by skipping every test and turning our whole API surface into one big no-op. Torch is a huge dependency we probably shouldn't have _even if_ it supported Python 3.10.

> and not all the users need to run winograd test.

None of our _users_ need to run any test. They need to use the library. The developers and CI need to run the tests.

yamaguchi1024:
> None of our users need to run any test. They need to use the library. The developers and CI need to run the tests.

Sure. Ideally, we probably want to separate EXO dependency and EXO test dependency.

> Torch is a huge dependency we probably shouldn't have even if it supported Python 3.10.

I'm on board with that

yamaguchi1024:
This PR is ready for review @gilbo @alexreinking

gilbo (approved):

PR #165, opened 2022-02-24
Fix match_pattern to take alloc sizes into account

### LINKED ISSUE #139: Fix match_pattern
Currently, it can't differentiate `a : i8[_]` and `a : i8[N, M, K]`, where latter should be more specific.

yamaguchi1024:
Fixes #139


alexreinking (approved):
This looks great, thanks!

PR #180, opened 2022-04-05
Move gemmini code under platform

yamaguchi1024:


alexreinking:
I don't think `harness_gemmini.py` should be moved into the main sources? It certainly _can't_ be right now since it imports `pytest` and that's not listed in the package requirements, so doing this would break the package.

yamaguchi1024:
Removed the `pytest` import as it was not used anywhere. I'm not sure where harness should reside in, maybe not in the platform but some helper directory that users can have access to?

alexreinking:
> I'm not sure where harness should reside in, maybe not in the platform but some helper directory that users can have access to?

I'm not convinced users need access to our Gemmini test harness? I thought this was fine in `tests`.

yamaguchi1024:
I think we should have a helper function that generates the main function exposed to the user, because otherwise, it will be too demanding for novice users to "just run" Exo. I think we don't want a novice users finding Exo repository and trying to use it, and realize that they need to write the main function to call procedures by themselves and drop it.

alexreinking:
> I think we should have a helper function that generates the main function exposed to the user, because otherwise, it will be too demanding for novice users to "just run" Exo. I think we don't want a novice users finding Exo repository and trying to use it, and realize that they need to write the main function to call procedures by themselves and drop it.

I get your point and I generally agree, but I think this merits a much more careful design. Halide has similar features (RunGenMain) we should look at.

I don't think moving this file is necessary for the purposes of artifact evaluation (i.e. this PR), so let's revert that for now and we can discuss how to generate command line tools (and perhaps more! e.g. Python/Matlab/Julia bindings, etc.) later.

yamaguchi1024:
I didn't know about RunGenMain, that sounds super relevant!
Sounds good, let's discuss about this more carefully. I'll remove the harness.

tests/gemmini/harness_gemmini.py
```diff
@@ -1,17 +1,13 @@
...
-
 from exo import compile_procs
 
 GEMMINI_ROOT = os.getenv('GEMMINI_ROOT')
 if GEMMINI_ROOT is None:
     RISCV = os.getenv('RISCV')
     if RISCV is None:
         pass
-        #pytest.skip("skipping gemmini tests; could not find chipyard",
-        #            allow_module_level=True)
```
alexreinking, line 14: Do we not actually want this here? Seems like the tests ought to skip if chipyard isn't available.

yamaguchi1024: It won't compile&run gemmini tests unless RISCV is present, because of this line:
https://github.com/ChezJrk/exo/blob/master/tests/gemmini/harness_gemmini.py#L247

Without RISCV environment, it'll just run the frontend.

alexreinking (approved):

gilbo:
Why does this need to go in platforms? Also, relatedly we want to remove platforms from the source ultimately, not add to it.

PR #184, opened 2022-04-11
Remove deprecated par loops from reorder and split

skeqiqevian:


gilbo (commented):
Was it necessary to change the tests in order for them to pass?

yamaguchi1024 (approved):

PR #186, opened 2022-04-13
Improve simpilfy to normalize polynomial before constant evaluation

yamaguchi1024:


src/exo/LoopIR_scheduling.py
```diff
@@ -2564,11 +2564,147 @@ def map_e(self, e):
...
+            new_hi = self.index_start(s.hi)
+            new_body = self.map_stmts(s.body)
+            return [type(s)(s.iter, new_hi, new_body, s.eff, s.srcinfo)]
+        elif isinstance(s, LoopIR.Alloc):
+            if isinstance(s.type, T.Tensor):
+                new_hi = [ self.index_start(e) for e in s.type.hi ]
+                new_typ = T.Tensor(new_hi, False, s.type.basetype())
+                return [LoopIR.Alloc(s.name, new_typ, s.mem, s.eff, s.srcinfo)]
+
+        return super().map_s(s)
```
gilbo, line 2697: Could you eliminate all of these different `map_*` cases by simply checking at the top of `map_e` whether or not an expression is of `index` or `size` type (i.e. is `indexable`) and then returning the result of `index_start` if it is?

src/exo/LoopIR_scheduling.py
```diff
@@ -2564,11 +2564,147 @@ def map_e(self, e):
...
+        elif isinstance(e, LoopIR.USub):
+            e_map = self.normalize_e(e.arg)
+            return { key: -e_map[key] for key in e_map }
+        elif isinstance(e, LoopIR.BinOp):
+            lhs_map = self.normalize_e(e.lhs)
+            rhs_map = self.normalize_e(e.rhs)
+            return self.concat_map(e.op, lhs_map, rhs_map)
+        else:
+            assert False, ("index_start should only be called by"+
+                           f" an indexing expression. e was {e}")
```
gilbo, line 2614: `ReadConfig` can have `indexable` type, and is unhandled.

src/exo/LoopIR_scheduling.py
```diff
@@ -2564,11 +2564,147 @@ def map_e(self, e):
...
+                return { key: lhs[key]*rhs[self.C] for key in lhs }
+            else:
+                assert self.C in lhs
+                return { key: rhs[key]*lhs[self.C] for key in rhs }
+        else:
+            assert False, "bad case"
+
+    def normalize_e(self, e):
+        if isinstance(e, LoopIR.Read):
+            assert e.type.is_indexable()
```
gilbo, line 2600: Should this assertion be lifted up to the top scope of this method, applicable to all cases?

src/exo/LoopIR_scheduling.py
```diff
@@ -2564,11 +2564,147 @@ def map_e(self, e):
...
 
+
+class _DoNormalize(LoopIR_Rewrite):
+    def __init__(self, proc):
+        self.C = Sym("temporary_constant_symbol")
+        super().__init__(proc)
+
+        self.proc = InferEffects(self.proc).result()
+
+    def concat_map(self, op, lhs, rhs):
```
gilbo, line 2575: It would be good to have a comment here or somewhere about what these `dict`s that we're operating on actually mean.

i.e. the dictionaries actually represent linear combinations of symbols, which are the keys.  The coefficient for a particular symbol is the value that key maps to.  One option for factoring this code is that you could choose to create an "affine expression" class that maintains the dictionary and implements the relevant operations.

src/exo/LoopIR_scheduling.py
```diff
@@ -2564,11 +2564,147 @@ def map_e(self, e):
...
+        if op == '+':
+            # if has same key: add value
+            common = { key: (lhs[key]+rhs[key]) for key in lhs if key in rhs }
+            return lhs | rhs | common
```
gilbo, line 2579: Are the semantics of union such that the right-hand side of the union is guaranteed to overwrite the left in the event of a key clash?

src/exo/LoopIR_scheduling.py
```diff
@@ -2564,11 +2564,147 @@ def map_e(self, e):
...
+            neg_rhs = { key: -rhs[key] for key in rhs }
+            return lhs | neg_rhs | common
+        elif op == '*':
+            # rhs or lhs NEEDS to be constant
+            assert len(rhs) == 1 or len(lhs) == 1
+            # multiply the other one's value by that constant
+            if len(rhs) == 1 and self.C in rhs:
+                return { key: lhs[key]*rhs[self.C] for key in lhs }
+            else:
+                assert self.C in lhs
```
gilbo, line 2593: `len(lhs) == 1 and self.C in lhs`

src/exo/LoopIR_scheduling.py
gilbo: If you do have a class, the class could also manage the conversion of itself back to `LoopIR.expr`s

src/exo/LoopIR_scheduling.py
```diff
@@ -2564,11 +2564,147 @@ def map_e(self, e):
...
+        
+        delete_zero = { key: n_map[key] for key in n_map if n_map[key] != 0 }
+        new_e = LoopIR.Const(0, T.int, e.srcinfo)
+        for key, val in delete_zero.items():
+            if val > 0:
+                # add
+                new_e = LoopIR.BinOp('+', new_e, get_loopir(key, val), e.type, e.srcinfo)
+            else:
+                # sub
+                new_e = LoopIR.BinOp('-', new_e, get_loopir(key, val), e.type, e.srcinfo)
```
gilbo, line 2661: Does this need to be `-val`?

src/exo/LoopIR_scheduling.py
```diff
@@ -2564,11 +2564,147 @@ def map_e(self, e):
...
+        def get_loopir(key, value):
+            vconst = LoopIR.Const(value, T.int, e.srcinfo)
+            if key == self.C:
+                return vconst
+            else:
+                readkey = LoopIR.Read(key, [], e.type, e.srcinfo)
+                return LoopIR.BinOp('*', vconst, readkey, e.type, e.srcinfo)
+        
+        delete_zero = { key: n_map[key] for key in n_map if n_map[key] != 0 }
+        new_e = LoopIR.Const(0, T.int, e.srcinfo)
```
gilbo, line 2654: Will this just generate gratuitous leading zeros everywhere, or does simplification take care of that for you?

yamaguchi1024: Simplification's constant folding path takes care of eliminating the leading zeros

src/exo/LoopIR_scheduling.py
yamaguchi1024: Yes!

src/exo/LoopIR_scheduling.py
yamaguchi1024: I'll add comments, thanks!

src/exo/LoopIR_scheduling.py
yamaguchi1024: Thanks for catching!

src/exo/LoopIR_scheduling.py
yamaguchi1024: Thanks for catching this!!

yamaguchi1024: Added a test for this

src/exo/LoopIR_scheduling.py
yamaguchi1024: Yes, thanks!

yamaguchi1024:
codecov is being silly

gilbo (approved):

PR #187, opened 2022-04-20
Cursors (w/o forwarding)

yamaguchi1024:
(This PR is by Alex and me)
On behalf of breaking up the cursor implementation to smaller PRs, I think the current state is a good time for an initial review because if we're implementing Alex's idea of scheduling operations returning an fwd lambda function (which I think is a good idea), it will require changes to all the scheduling operations.

This PR implements the Cursor object, cursor movements, and cursor retrieval from the string pattern.


src/exo/cursors.py
```diff
@@ -0,0 +1,228 @@
...
+class ExoIR(Enum):
+    Assign = 0
+    Reduce = 1
+    WriteConfig = 2
+    Pass = 3
+    If = 4
+    For = 5
+    Alloc = 6
+    Call = 7
+    WindowStmt = 8
```
gilbo, line 30: This seems redundant with `LoopIR` itself in a way that can fall out of sync.

src/exo/cursors.py
```diff
@@ -0,0 +1,228 @@
...
+from weakref import ReferenceType
+
+from .API_types import ProcedureBase
+from .LoopIR import LoopIR
+
+
+class CursorKind(Enum):
+    Node = 0  # Pointing directly at a node
+    # BlockFront = 1  # Gap before the first statement of a block
+    # BlockEnd = 2  # Gap after the last statement of a block
```
gilbo, line 16: Are sequences of statements not supported?

src/exo/cursors.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+class InvalidCursorError(Exception):
+    pass
+
+
+@dataclass
+class Cursor:
+    _proc: ReferenceType[ProcedureBase]
+    _node: ReferenceType[Union[LoopIR.proc, LoopIR.stmt, LoopIR.expr]]
```
gilbo, line 40: Why should a cursor ever be pointed at a `proc` rather than its body?

src/exo/cursors.py
```diff
@@ -0,0 +1,228 @@
...
+    # Static constructors
+    # ------------------------------------------------------------------------ #
+
+    @staticmethod
+    def root(proc):
+        return Cursor.from_node(proc, proc.INTERNAL_proc(), [])
+
+    @staticmethod
+    def from_node(proc, node, path):
+        return Cursor(weakref.ref(proc), weakref.ref(node), path)
```
gilbo, line 54: This probably ought not be exposed to users, but if `Cursor` objects are exposed to users, then we probably need to expose this class for typechecking.  Therefore a user would be able to call `Cursor.from_node`.

One solution to this problem might be to have a cursor wrapper class in `API.py` and repeat all the user-accessible methods there, wrapping this actual implementation.  Otherwise, we need to indicate that this class is exposed and be more disciplined about underscores on methods.

src/exo/cursors.py
gilbo: i.e. can we just use references to the classes themselves rather than an enumeration?

src/exo/cursors.py
gilbo: Secondly, looking at this implementation it seems like we only ever really need to have a `LoopIR.proc`, not a `Procedure` (which is the external wrapper object for the API)  Should we change what's stored here then?

gilbo (commented):
This is a very partial review based on reading through `cursors.py`.  I want to drop some comments ASAP in case they're useful.

Overall, I'm a bit concerned that this cursor implementation seems half complete.  i.e. it seems like there's a reasonable path punched through to get the old API to work.  That's great!  But gaps and statement blocks are still not really fleshed out.  Even if they aren't tested that well, it would be good to at least have a complete vision of how they should be represented here as part of this PR.  That should remain orthogonal to the cursor forwarding.  (Note: cursor navigation on a single procedure can be fleshed out without forwarding)

src/exo/cursors.py
yamaguchi1024: +1 for having a cursor wrapper class in API.py

src/exo/cursors.py
yamaguchi1024: A sequence of statements is currently represented as a list of Cursor objects.

src/exo/cursors.py
yamaguchi1024: This was patched in for the idea of making Cursor more like a QAST, to allow flexible user introspection, but is currently unused. Currently, `_node()` is a reference to a LoopIR, but we probably don't want users to introspect on LoopIR, so I guess we need something similar to QAST.

src/exo/API.py
```diff
@@ -200,14 +200,30 @@ def is_instr(self):
     def get_instr(self):
         return self._loopir_proc.instr
 
+    def find_cursor(self, pattern):
+        cursors = match_cursors(self, pattern, call_depth=1)
+        assert isinstance(cursors, list)
+        if not cursors:
+            raise SchedulingError('failed to find statement', pattern=pattern)
```
gilbo, line 207: Should say `failed to find cursor` or `failed to find matches` not statement specifically

src/exo/API.py
```diff
@@ -233,6 +249,9 @@ def get_ast(self, pattern=None):
                 assert len(results) == 1
                 return results[0]
 
+    def body(self):
```
gilbo, line 252: rename to `root`?

tests/test_config.py
```diff
@@ -349,7 +349,10 @@ def ld_i8(
     ld_i8 = ld_i8.configwrite_after('ConfigLoad.scale = _', ConfigLoad,
                                     'src_stride', 'stride(src, 0)')
     ld_i8 = ld_i8.replace(do_ld_i8, 'for i in _:_')
-    ld_i8 = ld_i8.replace(config_ld_i8, 'ConfigLoad.scale = scale')
+    ld_i8 = ld_i8.replace(
```
yamaguchi1024, line 352: Do we have a mechanism to specify how many statements we want the selection cursor to hold? For example, it'll be nice to write here `ld_i8.find_cursor('ConfigLoad.scale = _;', 2)` or sth like that.

alexreinking: Every cursor in the list returned by this call will be a 1-statement selection. However, the list might be of length greater than 1 if there are multiple matches. We might want to add a method to `Selection` to _extend_ the selection to include the next `n` statements. With what's here now you could write:

```
[sel[:2] for sel in ld_i8.find_cursor('ConfigLoad.scale = _ ; _')]
```

which would ensure each match had length at _most_ 2.

alexreinking: You could also imagine extending the pattern language with something like:

```
ld_i8.find_cursor('ConfigLoad.scale = _ ; _[1]')
```
where `_[1]` means "match exactly 1 statement". You could also do things like `_[:2]` to mean "match up to two" or `_[3:]` to mean "match at least three".

src/exo/API.py
```diff
@@ -200,14 +200,30 @@ def is_instr(self):
     def get_instr(self):
         return self._loopir_proc.instr
 
+    def find_cursor(self, pattern):
```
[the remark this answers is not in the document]
alexreinking: Renamed to `find_cursors`.

src/exo/API.py
alexreinking: Changed to say "matches" instead of "statement".

src/exo/API.py
alexreinking: Removed outright since it's untested/unused and pending further API discussion.

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+class Cursor(ABC):
+    _proc: ReferenceType[ProcedureBase]
+
+    # ------------------------------------------------------------------------ #
+    # Static constructors
+    # ------------------------------------------------------------------------ #
+
+    @staticmethod
+    def root(proc: ProcedureBase):
+        return Node(weakref.ref(proc), [])
```
yamaguchi1024, line 95: @alexreinking @gilbo I still think this function is slightly confusing, because "cursors" is an intuitive concept to users because it directly corresponds to a familiar cursor/selection concept in the text editor. "Having a root of the AST tree" is a real motivation for compiler writers, but for users what is the corresponding intuitive concept? proc? But cursors is not a proc. I'm starting to think what Gilbert was suggesting in the meeting about making root a selection of all statements makes sense.

tests/test_config.py
yamaguchi1024: I think your first idea is clearer and makes more sense, extending the string pattern matching doesn't sound like a good idea.

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+@dataclass
+class Selection(Cursor):
+    _path: list[tuple[str, Union[int, range]]]  # is range iff last entry
+
+    # ------------------------------------------------------------------------ #
+    # Navigation (implementation)
+    # ------------------------------------------------------------------------ #
+
+    def parent(self) -> Node:
+        return Node(self._proc, self._path[:-1])
```
yamaguchi1024, line 228: Is it guaranteed that `self._path` exists?

alexreinking: Yes. There are only three subtypes of `Cursor`, all internal. Users must not extend `Cursor`.

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+        for attr in args:
+            children = getattr(n, attr)
+            if isinstance(children, list):
+                for i in range(len(children)):
+                    yield self.child(attr, i)
+            else:
+                yield self.child(attr, None)
+
+    # ------------------------------------------------------------------------ #
+    # Navigation (block selectors)
```
yamaguchi1024, line 474: What were our plans on statements/expression introspection?
I thought there were two ways:
1. Implement introspection function for all cases in cursors object e.g., `.is_if()` `.is_loop()`
2. Implement QAST-ish class object exposed to users so that they can write `if isinstance(QAST.Seq, s)`

I guess we are going with the 1?

alexreinking: Not sure, we'll have to figure that out in a future PR. Right now I'm just looking at the `LoopIR` node directly via `isinstance` and `Node.node()`. I would be fine attaching introspection functions to cursors as in (1), to insulate the users from the underlying types. The `node()` function will have to be made private.

src/exo/cursors.py
alexreinking: The corresponding intuitive concept is the function AST. I think it makes absolutely no sense at all to restrict cursors from being able to point there. Take for example the following procedure:

```
def example(n: size, a: R[n]):
  assert n > 3
  a[0] = 1
  a[1] = 2
  a[2] = 3
  a[n-1] = 0
```

From an internal data structures perspective, preventing Cursors from being able to point at `n > 3` is unjustifiable to me, period. Doing so would strip cursors of their ability to find and replace all instances of `n` in the whole procedure. To implement such a directive, it would need to loop over each top-level statement and do the replacement separately, then re-write the signature and assertions using... bespoke code? A bespoke forwarding function that's unique to it? Why?

Scaled over an entire scheduling API, this is nonsense. It will lead to a huge amount of redundant work.

I can't justify it from an API perspective either. Imagine you have the above procedure in a text editor and your cursor is pointed at `a[0] = 1`. "What do you mean I can't go up?"

tests/test_config.py
alexreinking: A Window expression and a selection are entirely different concepts... I'm not sure how they would relate. The indexing operations Selection supports are basically the same as those of a Python list, you just can't step by anything other than 1.

src/exo/cursors.py
yamaguchi1024: I think having a function AST makes sense, but that should have its own class and methods and should be different from the class `Node`. Also I think the method name should be `func()` instead of `root()`, because `root()` is ambiguous.

Regarding pointing cursors to asserts, it is very unintuitive for users if they need to understand that "assertion is an annotation to procedures, not a statement". If we have `func()` or `root()`, users will need to do sth like:
```
c = foo.find_cursor("a[0] = 1")
assert = c.func().assert_stmts[0]
```
to get an assert statement, when they expect `c.prev()` would work.

tests/test_config.py
yamaguchi1024: Does `Selection` support range (e.g., `s[1:3]`)?

src/exo/cursors.py
yamaguchi1024: +1 for figuring out the details when we need this in the future

src/exo/cursors.py
alexreinking: > but that should have its own class and methods and should be different from the class Node.

Maybe as a subtype with fewer functions, but they fundamentally function the same and should share as much implementation as possible.

alexreinking: > it is very unintuitive for users if they need to understand that "assertion is an annotation to procedures, not a statement".

I think the fact that assertions _aren't_ statements in unintuitive :shrug:

alexreinking: But regardless, they could write `foo.find_cursor("assert e")` and the fact it's an assert PAST tells the matching system where to look.

tests/test_config.py
alexreinking: Yes, that statement returns a sub-selection containing the second and third statements of `s`, the same as a Python list.

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+                assert isinstance(cursor, Selection)
+                idx = fwd_sel(old_idx)
+
+            return evolve(old_path[:depth - 1] + [(attr, idx)] + old_path[depth:])
+
+        return forward
+
+
+@dataclass
+class Selection(Cursor):
```
gilbo, line 220: Anything pointed to by a class is selected by the cursor.  If this means a sequence of statements, it should probably be called a sequence of statements, not a selection

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+
+    def __len__(self):
+        _, _range = self._path[-1]
+        return len(_range)
+
+    # ------------------------------------------------------------------------ #
+    # AST mutation
+    # ------------------------------------------------------------------------ #
+
+    def replace(self, stmts: list):
```
gilbo, line 282: Is this public?

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+            #     del_range.stop if rng.stop in del_range[1:] else rng.stop
+            # )
+
+            start = rng.start + n_diff * (rng.start >= del_range.stop)
+            stop = rng.stop + n_diff * (rng.stop >= del_range.stop)
+            return range(start, stop)
+
+        return self._make_forward(new_proc, fwd_node, fwd_gap, fwd_sel)
+
+    def delete(self):
```
gilbo, line 331: Another user visible function?

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+    def _select_attr(self, attr: str):
+        stmts = getattr(self.node(), attr)
+        assert isinstance(stmts, list)
+        return Selection(self._proc, self._path + [(attr, range(len(stmts)))])
+
+    # ------------------------------------------------------------------------ #
+    # Conversions
+    # ------------------------------------------------------------------------ #
+
+    def select(self) -> Selection:
```
gilbo, line 492: Avoid "selection" name

gilbo: Maybe "selections" can be renamed as "block" or "statements"?

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+        attr, i = self._path[-1]
+        if i is None:
+            raise InvalidCursorError('cannot select nodes outside of a block')
+        return Selection(self._proc, self._path[:-1] + [(attr, range(i, i + 1))])
+
+    # ------------------------------------------------------------------------ #
+    # AST mutation
+    # ------------------------------------------------------------------------ #
+
+    def replace(self, ast):
```
gilbo, line 502: public?

src/exo/cursors.py
```diff
@@ -0,0 +1,599 @@
...
+        return self._translate(Gap, self._path, -dist)
+
+    def next(self, dist=1) -> Gap:
+        return self._translate(Gap, self._path, dist)
+
+    # ------------------------------------------------------------------------ #
+    # AST mutation
+    # ------------------------------------------------------------------------ #
+
+    def insert(self, stmts: list[LoopIR.stmt], policy=ForwardingPolicy.AnchorPre):
```
gilbo, line 559: public?

gilbo (commented):
I'm leaving a couple of surface level comments here.

I'm still a bit confused overall by whether all these things in the class hierarchy are intended to be a user-visible API or not.  I think we need to sort out the issue of separating the user-visible portion of the API from the implementation before checking in this PR.  We can delay the design discussions about the API, but it should be defined by an explicit wrapper class.

If I didn't find a number of examples of accidentally exposed, compiler-internal functionality in this review I might not have felt as strongly.  But given I did find those, I think separating the code is the best way to prevent goofs going forward.

tests/test_cursors.py
```diff
@@ -0,0 +1,692 @@
...
+def bar(n: size, m: size):
+    x: f32
+    for i in par(0, n):
+        for j in par(0, m):
+            x = 0.0
+            x = 1.0
+            x = 2.0
+            x = 3.0
+            x = 4.0
+            x = 5.0
```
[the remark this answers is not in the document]
alexreinking: This cuts down dramatically on code duplication, though. I can make a fixture to keep it from being top-level.

src/exo/cursors.py
alexreinking: I'll go with Block

alexreinking: Done.

alexreinking: Done.

src/exo/cursors.py
alexreinking: Renamed to "Block"

tests/test_cursors.py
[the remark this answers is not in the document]
alexreinking: Made into module-scoped fixture

alexreinking:
> If I didn't find a number of examples of accidentally exposed, compiler-internal functionality in this review I might not have felt as strongly.

IIRC we wanted users to be able to build their own scheduling directives. Was the idea that (1) user-defined directives should be built _only_ out of the verified building blocks that we provide, or (2) that users can _additionally_ write arbitrary transformations that they verify themselves?

I understood (2), so this wasn't an accident or a goof. I thought it would be helpful to expose these (granted, I should probably have named them like `unsafe_insert`, etc.). As I gather from your objection, my understanding was incorrect; (1) was intended. **That's totally fine.** One resolution could be to mark them private as in `_delete`, `_replace`, and `_insert`. Happy to make that change. :slightly_smiling_face: 

>  the API, but it should be defined by an explicit wrapper class. [...] But given I did find those, I think separating the code is the best way to prevent goofs going forward.

I want to _gently_ push back on this. :peace_symbol:

I think trying to hide symbols, classes, etc. in Python is noble, but impractical. Unlike other popular OOP systems, Python provides no clean mechanisms to do this. There are only civil agreements between library authors and their users, like the leading-underscore convention, and explicit documentation.

If we were writing C# code, we could protect `replace`, `delete`, and `insert` by marking them `internal` and then expose them to our test module via the `InternalsVisibleTo` attribute. In Java, we could probably get away with making them package-private. And so on in other languages.

But in Python, we can only make it marginally more challenging to break our rules. We can add a wrapper, but it's not free. Python doesn't really try to optimize very much, so it costs both memory and runtime. These overheads can be surprisingly large. It also adds a good bit of redundant code (which is annoying to maintain), or introspection-y code (which is slow). Despite this, determined users will still plow through to the private, wrapped, object and call private methods if they feel like it. Hence, wrappers that simply hide things tend to feel like foot-guns to me.

gilbo:
Yes, users should only be able to build transforms out of certified primitives. We have an unsafe “assert equivalent” escape hatch, but that requires the user to build the second procedure ex nihilo. We could choose to provide code quoting/assembly metaprogramming primitives in the future, which in combination with reflection and asserting equivalence could be used to implement arbitrary transforms, but we would virtually guarantee no-one does that by making it that difficult.

One consequence or alternate way to think about the above is that we ought to be able to produce proof collateral from Exo certifying the equivalence of any chain of transformations, even in the presence of abstracted scheduling operations.

alexreinking:
> Yes, users should only be able to build transforms out of certified primitives.

Gotcha. Will mark those methods as private.

alexreinking:
Closing in favor of #193 to correct commit authorship.

PR #189, opened 2022-05-10
Refactoring the API as part of Effect-Check/API audit

gilbo:
These should all be extremely trivial changes; just trying to separate from more meaningful changes in a later PR.


alexreinking (approved):

yamaguchi1024:
@gilbo Did `data_reuse` and `delete_config` changed their locations? What was the motivation for that?

gilbo:
I can’t recall, but I think the order made less sense to me. We’ll want to organize this better in the future regardless.

yamaguchi1024:
@gilbo Sure, this change looks good to me!

PR #193, opened 2022-05-18
Add initial cursor skeleton

alexreinking:
This PR lays the foundation for building a new scheduling API on top of our cursors idea. In particular, this PR includes:

1. A new abstract `Cursor` type that has three concrete subtypes: a `Node`, a `Block`, and a `Gap`.
2. Public navigation methods on these objects to help select parents, children, and gaps surrounding nodes and blocks.
3. _Internal_ functions for making unsafe AST changes that produce a forwarding function as collateral.
4. Modifications to the pattern matching system that return cursors instead of (lists of) raw LoopIR
5. A bugfix in the pattern matching system. Matching a single statement now always returns a block containing that single statement, rather than the statement and the remainder of the block.

Although cursors are exposed through the public API, they are currently not very useful. Only the navigation functions are public, as is the underlying proc getter. Otherwise, no other public APIs accept them.

This type will ultimately serve (at least) three purposes:

1. It is a forwardable handle for scheduling directives, such that our vision of user-defined scheduling directives may be realized.
2. It is an introspection tool for users to make particular scheduling decisions as in (1)
3. It may be used _internally_ for efficient AST modification and simplified forwarding definitions.

No progress is made on (2) in this PR.


src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+                n = n[idx]
+        return n
+
+    def _rewrite_node(self, fn):
+        """
+        Applies `fn` to the AST node containing the cursor. The callback gets
+        the following arguments:
+        1. The parent node
+        2. The relevant list of children
+        3. The last step in the path (the attribute and index info)
```
gilbo, line 158: This seems to not be true?  Below `fn()` only gets a single node.  (The parent?)

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+        return impl(self.proc().INTERNAL_proc(), self._path)
+
+    def _make_forward(self, new_proc, fwd_node, fwd_gap, fwd_sel):
+        orig_proc = self._proc
+        depth = len(self._path)
+        attr = self._path[depth - 1][0]
+
+        def forward(cursor: Cursor) -> Cursor:
+            if cursor._proc != orig_proc:
+                raise InvalidCursorError('cannot forward unknown procs')
```
gilbo, line 187: Should this be an `assert` error (i.e. internal) or can a user trigger this?

alexreinking: A user can trigger this by trying to use a cursor derived from one proc in a scheduling directive applied to a different proc.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+    def after(self, dist=1) -> Gap:
+        attr, _range = self._path[-1]
+        assert len(_range) > 0
+        return self.parent().child(attr, _range.stop - 1).after(dist)
+
+    def prev(self, dist=1) -> Cursor:
+        # TODO: what should this mean?
+        #  1. The node after the block?
+        #  2. The block shifted over?
+        #  3. The block of nodes leading up to the start?
```
gilbo, line 243: One possibility for `prev` and `next` in general with statements is to have something like
`c.next(range(3))`
to mean get a block of the next 3 statements after this statement or after this block?  (I'm not sure I like that syntax I just put up, but just to jot down the idea)

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+    # Navigation (implementation)
+    # ------------------------------------------------------------------------ #
+
+    def parent(self) -> Node:
+        return Node(self._proc, self._path[:-1])
+
+    def before(self, dist=1) -> Gap:
+        attr, _range = self._path[-1]
+        assert len(_range) > 0
+        return self.parent().child(attr, _range.start).before(dist)
```
gilbo, line 232: We should probably implement these without the `parent().child(...)` pattern?
e.g. can't we just directly update the path to construct the `start` or `stop` node in the block, and then call `before` or `after`?

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+
+    def __getitem__(self, i):
+        attr, r = self._path[-1]
+        r = r[i]
+        if isinstance(r, range):
+            if r.step != 1:
+                raise IndexError('block cursors must be contiguous')
+            return Block(self._proc, self._path[:-1] + [(attr, r)])
+        else:
+            return self.parent().child(attr, r)
```
gilbo, line 271: The range tricks here are pretty cool!

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+        This is an UNSAFE internal function for replacing a block in an AST with
+        a list of statements and providing a forwarding function as collateral.
+        It is meant to be package-private, not class-private, so it may be
+        called from other internal classes and modules, but not from end-user
+        code.
+        """
+        assert self._path
+        assert len(self) > 0
+
+        def update(node):
```
gilbo, line 292: maybe call this argument `parent` instead of `node` to make its role easier to guess?

gilbo: Might be good to similarly rename some other local variables named `node`.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+            attr, i = self._path[-1]
+            children = getattr(node, attr)
+            new_children = children[:i.start] + nodes + children[i.stop:]
+            new_children = new_children or empty_default or []
+            return node.update(**{attr: new_children})
+
+        from .API import Procedure
```
gilbo, line 299: Would it be better to `from . import API` at the top of this file and then just call `API.Procedure` below?

src/exo/cursors.py
alexreinking: Ah, yes, this comment is stale. I removed the latter two arguments after I found they weren't useful. Only the parent node is passed. Thanks for catching this.

src/exo/cursors.py
gilbo: btw, if you've never read Craig Chambers' papers on efficient implementation of the language `Self`, you might find those interesting w.r.t. these kinds of issues in dynamic language implementation

src/exo/cursors.py
alexreinking: Yes, that's a good idea. I'll make that change before merging.

src/exo/cursors.py
alexreinking: IIRC it was causing a circular import error, but that might have been fixed now that `API` doesn't import `cursors` anymore.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+        p = Procedure(self._rewrite_node(update))
+
+        return p, self._forward_replace(weakref.ref(p), len(nodes))
+
+    def _forward_replace(self, new_proc, n_ins):
+        del_range = self._path[-1][1]
```
gilbo, line 305: Does
`_, del_range = self._path[-1]`
work?  If so, it might be more consistent with your style elsewhere in this file.

alexreinking: It does, and will fix.

src/exo/cursors.py
gilbo: Maybe it was `from .API import Procedure` at the top level?  That can often cause circular import errors.  As a first recourse, I recommend switching to `from . import API` since simply importing the module will much more rarely cause the circular import problems.

src/exo/cursors.py
alexreinking: Going through `.child()` ensures that an IndexError is raised if you go too far.

src/exo/cursors.py
alexreinking: Thanks! :slightly_smiling_face:

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+        n_diff = n_ins - len(del_range)
+
+        def fwd_node(i):
+            if i in del_range:
+                raise InvalidCursorError('node no longer exists')
+            return i + n_diff * (i >= del_range.stop)
```
gilbo, line 311: This is overly strong when the cursor points to an invariant subtree.  i.e. the assumption here is that we don't use `replace` to replace something like `s1;s2` with `s2;s1` or even `s` with `if c: s`.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+
+        def fwd_gap(i):
+            if i in del_range[1:]:
+                raise InvalidCursorError('gap no longer exists')
+            return i + n_diff * (i >= del_range.stop)
```
gilbo, line 316: Am I right in understanding the policy here as "gaps sitting immediately before or immediately after the replaced code should remain valid"?

src/exo/cursors.py
alexreinking: Nope, I still get an error here:
```
tests/conftest.py:19: in <module>
    from exo import Procedure, compile_procs
src/exo/__init__.py:2: in <module>
    from .API import Procedure, compile_procs, proc, instr, config
src/exo/API.py:19: in <module>
    from .pattern_match import match_pattern, get_match_no, match_cursors
src/exo/pattern_match.py:14: in <module>
    from .cursors import Cursor, Node, Block
src/exo/cursors.py:11: in <module>
    from .API import Procedure
E   ImportError: cannot import name 'Procedure' from partially initialized module 'exo.API' (most likely due to a circular import) (/home/alex/Development/exo/src/exo/API.py)
```

src/exo/cursors.py
gilbo: That should be caught and signaled by the provenance tracking code then, no?

gilbo: i.e. some other (internal) code is going to have to find the right sequence of forwarding functions to apply anyway, and if it applies the wrong forwarding function then that would be an error in that other internal code, not a user error.

src/exo/cursors.py
gilbo: Got it

src/exo/cursors.py
alexreinking: Yeah, we'll need something slightly different like `._wrap` for a block-wrapping replacement. Or maybe something smarter that I'm not thinking of? The semantics of `._replace` here are really meant to be "this is totally new code" like what happens if you use the `.replace(instr, block)` directive or if you do something dramatic in the simplifier, etc.

src/exo/cursors.py
gilbo: Wait, it's impossible to go to far for `before()` or `after()`...

src/exo/cursors.py
alexreinking: Yes, a gap before/after a block that is replaced ends up before/after its replacement.

src/exo/cursors.py
gilbo: Yes, you're trying to do `from .API import Procedure`.  I'm saying, yes don't do that; it's not safe circularly.  Instead do
`from . import API`
at the top level, rather than putting an import statement in the middle of a method.

src/exo/cursors.py
alexreinking: I could imagine that this `._replace` ends up being a special case of `._wrap` where whatever tells the forwarding function where the old nodes went is `None` or something.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+            attr, i = self._path[-1]
+            children = getattr(node, attr)
+            new_children = children[:i.start] + nodes + children[i.stop:]
+            new_children = new_children or empty_default or []
+            return node.update(**{attr: new_children})
+
+        from .API import Procedure
+        p = Procedure(self._rewrite_node(update))
+
+        return p, self._forward_replace(weakref.ref(p), len(nodes))
```
gilbo, line 302: It seems that `len(nodes)` will give the wrong value when the `empty_default` is `pass_stmt`.  Have you thought carefully about this edge case?

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+    # ------------------------------------------------------------------------ #
+
+    def _node(self):
+        """
+        Gets the raw underlying node that's pointed-to. This is meant to be
+        compiler-internal, not class-private, so other parts of the compiler
+        may call this, while users should not.
+        """
+        if (n := self._node_ref()) is None:
+            raise InvalidCursorError('underlying node was destroyed')
```
gilbo, line 365: Should this be an internal assertion error?  If you intend for failed forwarding functions to raise `InvalidCursorError` eagerly during forwarding, then the only way you could trip this case is if the implementation of some forwarding function was broken.

gilbo: This seems even more pressing if I think about this method being called all over the `LoopIR_scheduling.py` code.  By the time we're doing things there, we probably oughtn't be triggering any more Cursor Invalidations

src/exo/cursors.py
gilbo: Oh, I see if dist is not 1, nm.

src/exo/cursors.py
gilbo: Ahh, so here's a thought.  Suppose that your general replacement (or forwarding?) call doesn't just accept a cursor defining the context (i.e. `C[•]`) in which the rewriting is happening, but also accepts a set of cursors to the "invariant sub-trees" along with cursors/paths to their new locations?  (maybe there's a way to be slightly more concise?)  If that were the case, then we would do the current checks (is the cursor being forwarded pointed at some context higher up in the tree?  Then just forward it trivially) but also do checks w.r.t. sub-trees.  (i.e. if the cursor is pointing at some node within an invariant sub-tree, then modify the pre-sub-tree prefix of the path to account for the relocation of the sub-tree.

gilbo: Note that in this view of things, gaps "in sub-trees" only get forwarded if they're properly _inside_ of the subtree, so in the case of `C[ s1 ; s2 ]`, the gap between `s1` and `s2` is neither on the edge of the context hole (handled below) or properly inside of the `s1` or `s2` subtrees.  Therefore it would be invalidated.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+
+    def next(self, dist=1) -> Node:
+        return self._translate(Node, self._path, dist)
+
+    # ------------------------------------------------------------------------ #
+    # Navigation (children)
+    # ------------------------------------------------------------------------ #
+
+    def child(self, attr, i=None) -> Node:
+        _node = getattr(self._node(), attr)
```
gilbo, line 400: If `child` is publicly visible, we should at least throw some kind of error (maybe not a cursor invalidation error?) when `getattr` fails?  Or I guess we're simply ok with the built-in error from `getattr`?

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+            yield from self._children_from_attrs(n, 'idx')
+        elif isinstance(n, (LoopIR.Const, LoopIR.WindowExpr, LoopIR.StrideExpr,
+                            LoopIR.ReadConfig)):
+            yield from []
+        elif isinstance(n, LoopIR.USub):
+            yield from self._children_from_attrs(n, 'arg')
+        elif isinstance(n, LoopIR.BinOp):
+            yield from self._children_from_attrs(n, 'lhs', 'rhs')
+        elif isinstance(n, LoopIR.BuiltIn):
+            yield from self._children_from_attrs(n, 'args')
```
gilbo, line 440: There should be an `else: assert False, f"case {type(n)} unsupported"` or something like that

src/exo/cursors.py
alexreinking: As we build it out, checks in that part of the system might supersede checks here. Then this could become a true assertion.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+        This is an UNSAFE internal function for replacing a node in an AST with
+        either a list of statements or another single node and providing a
+        forwarding function as collateral. It is meant to be package-private,
+        not class-private, so it may be called from other internal classes and
+        modules, but not from end-user code.
+        """
+        if self._path[-1][1] is not None:
+            # noinspection PyProtectedMember
+            # delegate block replacement to the Block class
+            return self.as_block()._replace(ast)
```
gilbo, line 491: Should we assert before delegating that `ast` is a list?  And below, we should assert that `ast` is not a list presumably...

src/exo/cursors.py
alexreinking: Doh, that fixes it, thanks

src/exo/cursors.py
alexreinking: Yeah, that does seem like a natural way of generalizing this. Lots more code, though, Will keep in mind for a near-future PR.

src/exo/cursors.py
alexreinking: As of right now, it's unreachable because the provenance system keeps everything alive. I think this will be worth thinking about more once that's fixed and this becomes reachable again.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+        from .API import Procedure
+        p = Procedure(self._rewrite_node(update))
+
+        return p, self._forward_replace(weakref.ref(p))
+
+    def _forward_replace(self, new_proc):
+        assert self._path[-1][1] is None
+
+        def fwd_node(_):
+            raise InvalidCursorError('cannot forward replaced nodes')
```
gilbo, line 507: As mentioned before, this is not necessarily true when sub-trees are involved.  This might also not be true when you have a node in the middle of a block being replaced.  A good sanity check is that if the `_replace` method above delegates to the `Block` form of `_replace`, does the forwarding policy similarly delegate?

src/exo/cursors.py
alexreinking: I'm making `_child` private.

src/exo/cursors.py
alexreinking: Fixed.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+        return self._translate(Gap, self._path, 1 - dist)
+
+    def after(self, dist=1) -> Gap:
+        return self._translate(Gap, self._path, dist)
+
+    def prev(self, dist=1) -> Node:
+        return self._translate(Node, self._path, -dist)
+
+    def next(self, dist=1) -> Node:
+        return self._translate(Node, self._path, dist)
```
gilbo, line 393: There is no bounds checking on any of these methods, or on the corresponding Gap methods.  Also note that bounds checking for nodes vs. Gaps is different by 1

src/exo/cursors.py
alexreinking: Yes, that can only happen when everything in the block is destroyed, so everything will be invalidated before any index math within the block would happne.

src/exo/cursors.py
```diff
@@ -0,0 +1,591 @@
...
+            def fwd_gap(i):
+                return i + ins_len * (i > ins_idx)
+        elif policy == ForwardingPolicy.AnchorPost:
+            def fwd_gap(i):
+                return i + ins_len * (i >= ins_idx)
+        else:
+            def fwd_gap(i):
+                if i == ins_idx:
+                    raise InvalidCursorError('insertion gap was invalidated')
+                return i + ins_len * (i > ins_idx)
```
gilbo, line 583: I think it would be ok to remove this policy knob and just have the forwarding behavior in the event of an insertion always be to invalidate the cursor.

src/exo/cursors.py
alexreinking: Added assertions

src/exo/cursors.py
alexreinking: I noticed this mid-review and have fixed it locally :upside_down_face:

src/exo/cursors.py
alexreinking: I'm imagining it will be useful to use the same gap to insert multiple things in a loop... forwarding the gap to after the insertion seems useful. :shrug:

src/exo/cursors.py
alexreinking: It's impossible _for now_. Eventually we will want users to know that they've attempted to use a cursor for a procedure that no longer exists... right?

alexreinking: Like, a user gets a cursor into a proc for scheduling it, makes a bunch of changes, then calls `.partial_eval` to make a bunch of variants and lets the original proc die... cursors don't keep procs alive, so it would die here.

PR #195, opened 2022-05-19
Deprecate add_guard

yamaguchi1024:


yamaguchi1024:
@gilbo Can you take a brief look at the changes, before I change all the use cite of `add_guard` ?

alexreinking (dismissed):

src/exo/API.py
```diff
@@ -679,17 +679,19 @@ def fuse_if(self, if1, if2):
...
-    def add_loop(self, stmt, var, hi):
+    def add_loop(self, stmt, var, hi, guard=False):
         if not isinstance(stmt, str):
             raise TypeError("expected first arg to be a string")
         if not isinstance(var, str):
             raise TypeError("expected second arg to be a string")
         if not isinstance(hi, int):
             raise TypeError("currently, only constant bound is supported")
+        if not isinstance(guard, bool):
+            raise TypeError("guard needs to be True or False")
```
alexreinking, line 690: Consider adding a test that tries setting `guard` to a non-bool.

alexreinking:
Looks great!

yamaguchi1024:
@alexreinking Thanks, I still need to remove more use cite of `add_guard` for this PR

alexreinking:
I think it's fine to merge this first, but if you want to wait, that's fine, too.

yamaguchi1024:
@gilbo @alexreinking I'll merge this PR if it looks alright~

tests/gemmini/conv/test_gemmini_conv.py
```diff
@@ -1,448 +0,0 @@
-from __future__ import annotations
```
alexreinking, line 1: Why was this file deleted?

tests/gemmini/conv/test_gemmini_conv_stride1.py
```diff
@@ -262,124 +304,6 @@ def test_conv_3():
 
     print(conv)
 """
```
alexreinking, line 306: Is this just an empty string now? If so, get rid of the string.

src/exo/API.py
```diff
@@ -696,17 +672,19 @@ def fuse_if(self, if1, if2):
 
         return Procedure(loopir, _provenance_eq_Procedure=self)
 
-    def add_loop(self, stmt, var, hi):
+    def add_loop(self, stmt, var, hi, guard=False):
```
alexreinking, line 675: At every call site, we're making sure to write `guard=True/False`. We should consider _requiring_ the use of the keyword by changing the signature as follows:

```suggestion
    def add_loop(self, stmt, var, hi, *, guard=False):
```
Arguments after the `*` are _keyword-only_.

alexreinking (changes requested):
Looks pretty good, just a little clean-up and I'll approve.

tests/gemmini/conv/test_gemmini_conv.py
yamaguchi1024: This is a super outdated test that was skipped (see line 8). It's fine not to delete this file, but it doesn't matter

alexreinking: Ok, I was just making sure it was intentional.

alexreinking (approved):
Looks good to me! Just did the merge with `master` myself, so we'll just wait for green. Go ahead and enable auto-merge.

PR #203, opened 2022-06-24
Accomodate Seq in bind_expr, fission_after, add_loop, double fission

yamaguchi1024:


src/exo/LoopIR_scheduling.py
```diff
@@ -1092,7 +1092,7 @@ def map_s(self, s):
 
         if isinstance(s, (LoopIR.ForAll, LoopIR.Seq)):
             body = self.process_block(s.body)
-            return [LoopIR.ForAll(s.iter, s.hi, body, s.eff, s.srcinfo)]
+            return [type(s)(s.iter, s.hi, body, s.eff, s.srcinfo)]
```
alexreinking, line 1095: Use `s.update(body=body)`

alexreinking (changes requested):

alexreinking:
We have a helper for this. Might as well use it.

src/exo/LoopIR_scheduling.py
```diff
@@ -2154,17 +2154,17 @@ def map_s(self, s):
                 # body doesn't depend on the loop
                 # and the body is idempotent
                 if s.iter in _FV(pre) or not _is_idempotent(pre):
-                    pre     = [LoopIR.ForAll(s.iter, s.hi, pre, None, s.srcinfo)]
+                    pre     = [type(s)(s.iter, s.hi, pre, None, s.srcinfo)]
```
alexreinking, line 2157: `s.update(body=pre, eff=None)`

yamaguchi1024:
Hmm isn't this okay? The helper is not used in other parts of this file

alexreinking:
> Hmm isn't this okay? The helper is not used in other parts of this file

I didn't retroactively replace all of these with the helper, but that doesn't mean we shouldn't use it in new code.

yamaguchi1024:
I'll fix this ASAP!

alexreinking (approved):

PR #208, opened 2022-06-28
Separate malloc definition and forward declarations

yamaguchi1024:


tests/conftest.py
```diff
@@ -178,6 +178,8 @@ def compile(
             procs: Union[Procedure, List[Procedure]],
             *,
             test_files: Optional[Dict[str, str]] = None,
+            include_dir = None,
+            additional_file = None,
```
yamaguchi1024, line 182: @alexreinking you might have a better idea than this?

src/exo/libs/gemm_acc_malloc.h
```diff
@@ -0,0 +1,11 @@
+#define HEAP_SIZE 100000
+#define DIM 16
```
gilbo, line 2: If these constants are not needed by calling code, should they instead be defined in the `.c` file?

src/exo/libs/gemm_malloc.h
```diff
@@ -0,0 +1,11 @@
+#define HEAP_SIZE 100000
+#define DIM 16
```
gilbo, line 2: ditto from other malloc

src/exo/libs/malloc.h
```diff
@@ -0,0 +1,9 @@
+#define HEAP_SIZE 100000
```
gilbo, line 1: ditto from other malloc

src/exo/libs/gemm_acc_malloc.h
gilbo: Note that there are ways to allow for these values to be overridden using compiler flags.  e.g. it may be better to write
```
#ifndef HEAP_SIZE
#define HEAP_SIZE 100000
#endif
```

Furthermore, you could run into trouble with name clashes given how generic these names are; e.g. `GEMMINI_ACC_HEAP_SIZE` would be a safer name?
(might be worth getting @alexreinking 's input on this)

gilbo (commented):
I added some basic comments; I did not review the testing infrastructure stuff and will leave that to @alexreinking to comment on, because he has more of a vision about that.

src/exo/libs/gemm_acc_malloc.h
yamaguchi1024: That's true, we never really need to change these parameters

yamaguchi1024:
@alexreinking Any thoughts on this?

src/exo/libs/malloc.c
```diff
@@ -1,12 +1,13 @@
-// clang-format off
-// these are filled in by Python's str.format()
-#define HEAP_SIZE {heap_size}
-// clang-format on
+#include "malloc.h"
```
alexreinking, line 1: Although it isn't reserved by the C standard, `malloc.h` is a header name used by many C standard library implementations, anyway, [including Apple's](https://opensource.apple.com/source/Libc/Libc-825.26/include/malloc/malloc.h.auto.html). Please rename it to something like `custom_malloc.h/c` to avoid conflicts.

tests/conftest.py
```diff
@@ -234,20 +236,31 @@ def _run_command(build_command, skip_on_fail):
         if skip:
             pytest.skip('Compile failure converted to skip')
 
-    def _generate_cml(self, test_files: Dict[str, str]):
+    def _generate_cml(self, test_files: Dict[str, str], include_dir=None, additional_file=None):
+        if additional_file:
+            lib_command = f'add_library({self.basename} "{self.basename}.c" "{additional_file}")'
+        else:
+            lib_command = f'add_library({self.basename} "{self.basename}.c")'
```
alexreinking, line 243:
```suggestion
        additional_file = f'"{additional_file}"' if additional_file else ""
```

alexreinking (changes requested):
This is really close, but needs a couple CMake tweaks, include guards, and a naming fix.

yamaguchi1024:
@alexreinking Addressed your comments, thanks!

alexreinking (approved):
Looks great, thanks!

PR #210, opened 2022-07-05
Enable global memory splicing in the end of C file

yamaguchi1024:


alexreinking:
This seems really suspect... what could go at the end of a source file like this that couldn't be moved to a separate compilation unit? One couldn't use any `static` methods defined here because they'd need forward declarations, and `static` forward declarations can't go in a header.

yamaguchi1024:
Hmmm I cannot show the actual example (it's Apple code)

yamaguchi1024:
The high level use case is when you have #ifdef SOME_MACRO in the
beginning, have the body code, and then #endif at the end of the file.

Well this requires compiler side of change, so..

yamaguchi1024:
https://gcc.gnu.org/onlinedocs/cpp/Ifdef.html
^ Description from gccc

alexreinking:
> https://gcc.gnu.org/onlinedocs/cpp/Ifdef.html 
> ^ Description from gcc

Yuka, I know what `#ifdef` does.

> The high level use case is when you have #ifdef SOME_MACRO in the beginning, have the body code, and then #endif at the end of the file.

Given the positioning of the would-be `#endif`, it sounds like you're trying to find a way to disable the entire compilation unit when `SOME_MACRO` is not defined. As in, the pair of `#if` and `#endif` surround every generated declaration. Is that correct?

yamaguchi1024:
Yes that's correct, and it seems that's a common approach when you support multiple targets and features in the system.

alexreinking:
> Yes that's correct, and it seems that's a common approach when you support multiple targets and features in the system.

Okay, this is something that is usually best handled in the build system... Xcode, for instance, has built-in ways of conditionally including source files: https://davedelong.com/blog/2018/07/25/conditional-compilation-in-swift-part-2/

In the worst case, you could create a wrapper C file that does something like:

```
#if defined(SOME_MACRO)
#include "path/to/generated/code.c"
#else
...
#endif
```
and then add this wrapper to the build system.

yamaguchi1024:
Ok I'll try the wrapper C file

yamaguchi1024:
@alexreinking C wrapper seems to work, so I'm closing this PR. Thanks!

PR #213, opened 2022-07-14
modified tests to run scheduling on platforms not supporting execution

gilbo:
This will eliminate upload conflicts for me running on my mac without access to AVX2, AVX512, or AMX.  I can at least catch scheduling errors this way and fix the tests without triggering failed CI tests.

### DIFF
--- a/tests/amx/test_amx_instr.py
+++ b/tests/amx/test_amx_instr.py
@@ -274,7 +274,8 @@ def test_matmul_on_amx_by_hand_i8(compiler, sde64):
         matmul_algorithm_i8(), get_transform_memory_i8(), matmul_on_amx
     ], t)
 
-def test_matmul_on_amx_scheduled_i8(compiler, sde64):
+@pytest.fixture
+def matmul_i8():
     size1 = 256
     size2 = 256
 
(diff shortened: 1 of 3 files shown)


tests/amx/test_amx_instr.py
```diff
@@ -348,6 +349,12 @@ def test_matmul_on_amx_scheduled_i8(compiler, sde64):
     print("Final scheduled algorithm:")
     print(amx)
 
+    return amx
+
+def test_gen_matmul_i8_amx(matmul_i8):
```
yamaguchi1024, line 354: @gilbo Why do you need this?

gilbo: This ensures that the scheduling code will get executed on my machine even though I don't have AMX installed.  Then I can catch errors arising from the scheduling code iteslf, even if I can't run a compilation check to make sure the result is still correct.

yamaguchi1024: Ahh so it's like a wrapper around even though you don't have `compiler`

yamaguchi1024 (approved):
Thanks!

PR #218, opened 2022-07-15
Merge neon-example

yamaguchi1024:
- moved `neon_dev` under `apps`
- @gilbo Can you take a look at a small bugfix in `stage_mem` in `src/exo/LoopIR_scheduling.py` ? I think it was a typo but wanted to make sure


gilbo (approved):
Seems fine to me; thanks for catching the goofs.  I bet I have a version of this somewhere that fixes it, but ¯\_(ツ)_/¯

PR #220, opened 2022-07-23
Cursor API Step 1

gilbo:
This change switches over the entire scheduling API to use the
```
p = sched(p, ...)
```
convention instead of the
```
p = p.sched(...)
```
convention.

Additionally, the pattern arguments now accept either pattern strings or cursors.  An `API_Cursors.py` wrapper for the internal cursors has been created, which are the standard for interface use now.

* This PR does not resolve the problem of having cursor forwarding yet.  That will be a "Cursor API Step 2" change in the future.
* This PR has not yet made sure that any untested code is properly updated to use the new API conventions.
* This PR makes some substantial cleanups to parts of the GEMMINI code, but is not a full cleanup of that code yet.
* This PR will leave large chunks of code in `API.py` commented out but not deleted.  This is intentional.  It will help me more smoothly merge in other branches that have been on hold; it may help others merge any outstanding branches as well.  In particular, I need this to help me merge in all of the yet-to-be-committed fixes to the program analyses.


yamaguchi1024:
We don't intend to expose `@sched_op([GapCursorA])` decorators to users, right? I wonder if it should start with an underscore to ensure it's for internal use only. We have `@proc` and `@instr()` and users can reasonably mistaken that `@sched_op` is one of them.

src/exo/LoopIR_scheduling.py
```diff
@@ -2531,13 +2521,18 @@ def _make_closure(name, stmts, var_types):
 
 
 class _DoInsertPass(LoopIR_Rewrite):
-    def __init__(self, proc, stmt):
+    def __init__(self, proc, stmt, before=True):
```
yamaguchi1024, line 2524: I'm guessing this is taking `before=True` because `insert_pass()` takes gaps and statements? I think it is fair to say that statement inserting primitives such as "insert pass" and "config write at" only take gaps.

src/exo/LoopIR_scheduling.py
```diff
@@ -3484,6 +3479,7 @@ class Schedules:
     DoExtractMethod = _DoExtractMethod
     DoParToSeq = _DoParToSeq
     DoReorderStmt = _DoReorderStmt
+    DoConfigWrite = _ConfigWrite
     DoConfigWriteAfter = _ConfigWriteAfter
```
yamaguchi1024, line 3483: Shouldn't we remove ConfigWriteAfter?

src/exo/__init__.py
```diff
@@ -1,9 +1,12 @@
 from .API import Procedure, compile_procs, proc, instr, config
 from .LoopIR_scheduling import SchedulingError
+from .parse_fragment import ParseFragmentError
 from .configs import Config
 from .memory import Memory, DRAM
 from . import query_asts as QAST
 
+from . import stdlib
```
yamaguchi1024, line 8: I am really not sure if it's a good idea to name our standard library `stdlib`. Maybe just `primitives` or `ops` ? That's more consistent with e.g., `apps/` and `platforms/`, no?

src/exo/internal_cursors.py
```diff
@@ -271,6 +280,24 @@ def __len__(self):
         _, _range = self._path[-1]
         return len(_range)
 
+    # ------------------------------------------------------------------------ #
+    # Block-specific operations
+    # ------------------------------------------------------------------------ #
+
+    def expand(self, lo=None, hi=None):
```
yamaguchi1024, line 287: I think it'll be useful if we can do a syntax sugar on this s.t.
```
c = cblock.next()
cblock = cblock + c
```

src/exo/platforms/gemmini.py
```diff
@@ -2,143 +2,172 @@
 
 from exo import proc, instr, DRAM, config, QAST
 from exo.libs.memories import GEMM_SCRATCH, GEMM_ACCUM
+from exo.stdlib.scheduling import *
```
yamaguchi1024, line 5: Gemmini changes are great, thank you for looking into this!
I'll ask Jennifer to test this on AWS before merging alongside with gemmini-rocc-tests stuff

gilbo:
> We don't intend to expose `@sched_op([GapCursorA])` decorators to users, right? I wonder if it should start with an underscore to ensure it's for internal use only. We have `@proc` and `@instr()` and users can reasonably mistaken that `@sched_op` is one of them.

The scheduling operators are exposed in `src/exo/stdlib/scheduling.py`.  This file should never be imported by user-code.

src/exo/LoopIR_scheduling.py
gilbo: Right now nothing in `LoopIR_scheduling.py` is written to take cursors.  Therefore it's impossible to pass a gap to one of these scheduling passes.  In order to hack this situation for the time being, I'm using this additional `before` argument.  When hopefully we switch over to just passing a cursor down to these operations, then we can change this interface.

src/exo/LoopIR_scheduling.py
gilbo: Probably; I'm trying to minimize changes to `LoopIR_scheduling.py` in this PR in order to make it easier to merge with other concurrent PRs.

gilbo: i.e. we should probably not remove dead code so much right now.

src/exo/__init__.py
gilbo: The idea is that `stdlib` lives under `exo` (i.e. `exo.stdlib`) in the Python module namespace.  Ultimately, I expect `stdlib` to hold things like "standard memories" excluding `apps` and `platforms` which are not part of the standard `exo-lang` distribution on PyPI.  We can create a second PyPI distribution for things like `platforms` libraries.

src/exo/internal_cursors.py
gilbo: Yes, there may be better choices.  (I would use the union operator `a | b` for what you're suggesting, not `+`)  This "expand" method is just there so I could solve one particular issue and move forward.  Don't take it as a long term commitment for what the API should be.

src/exo/__init__.py
yamaguchi1024: Hmmm I think `stdlib` is not saying anything, shouldn't users be able to import DRAM etc. and all the scheduling primitives directly with `from exo import *` ? I just think `exo.stdlib` should be just as simple as `exo`. And we can have peripheral code such as `builtins` and `platforms` separately and be accessed with `exo.builtins`

src/exo/internal_cursors.py
yamaguchi1024: Make sense. Shouldn't it be `+` because it's like expanding a list? union feels like unordered.

src/exo/__init__.py
gilbo: So, there will be a lot of scheduling primitives, not all of which will be atomic (e.g. `repeat`) and I expect that list to grow somewhat.  Generally I would expect `from exo import *` to get you things like `proc` that you'll need to use regularly, and then you may want to specifically bind the namespace of scheduling operations to a shorthand like `S` or `XS` (exo scheduling?) via
```
from exo.stdlib import scheduling as S
```

gilbo: `builtins` may be worth having under `exo` directly instead of `stdlib`.  I'm not sure.  As far as `platforms`, that shouldn't even be under `exo` in the namespace at all, and ultimately shouldn't be packaged with `exo-lang` in PyPI.  We can create a separate `exo-lang-platforms` package if we want to have standard "platform support" with lower guarantees of quality than the core `exo-lang` distribution.

yamaguchi1024: Shouldn't it be `exo.scheduling` then? I'm not sure why having `stdlib` in the middle is useful

gilbo:
We can change this eventually.  Right now, I just wanted to separate it from the other things in `exo` which should not be included by user code.

src/exo/internal_cursors.py
gilbo: It's a join in an appropriate lattice of sequences really.  For example, suppose I have cursors to two statements with another statement in between.
```
s1 <- c1
s2
s3 <- c2
```

The smallest block containing the two statements with cursors (`s1` and `s3`) is a block of all 3 statements.  That's neither a concatenation nor union.  It's a join of "intervals."

src/exo/__init__.py
gilbo: My reply to this got scrambled.  My ultimate plan for this is to move most of the current contents of the `exo` directory into a subdirectory called something like `compiler`.  At that point it may make sense to move `scheduling` to `exo.scheduling` and likewise for other submodules that we want users to import.  However, right now, there would be no distinction between `exo.scheduling` and `exo.LoopIR` which definitely should not be imported by user code.

tests/test_schedules.py
```diff
@@ -427,6 +427,64 @@ def foo(n: size, x: i8):
...
+def test_divide_dim_1(golden):
+    @proc
+    def foo(n: size, m: size, A : R[n + m + 12]):
+        x : R[n, 12, m]
+        for i in seq(0, n):
+            for j in seq(0, 12):
+                for k in seq(0, m):
+                    x[i, j, k] = A[i + j + k]
+
+    foo = divide_dim(foo, 'x', 1, 4)
```
yamaguchi1024, line 439: What will happen if there are multiple 'x' buffers in this proc? It seems the API does not loop over multiple Syms?

gilbo: All of the scheduling operations are now designed to apply exactly once.  Handling multiple applications of a scheduling operation should now be handled using higher-order combinators.  This will be even easier once step 2 of the API update is done, since then you can do something like `p.find_all(pattern)` to get a list of cursors, and then repeat an operation for each cursor in the list.

yamaguchi1024 (approved):

PR #221, opened 2022-07-28
remove loop if the body got emptied by assert_if

yamaguchi1024:


alexreinking (approved):
It would be nice to have some standard logic that enforced this.

yamaguchi1024:
@alexreinking yeah, I even think this should be part of original LoopIR_Rewrite class

gilbo:
It would be reasonable to subclass Rewrite for scheduling rewrite passes specifically. Doing so could help standardize passes and give a place to install helper functionality

yamaguchi1024:
Yeah that sounds good

PR #224, opened 2022-08-03
Implement commute()

yamaguchi1024:


src/exo/API_scheduling.py
```diff
@@ -715,6 +715,39 @@ def reorder_stmts(proc, block_cursor):
     loopir  = Schedules.DoReorderStmt(loopir, s1, s2).result()
     return Procedure(loopir, _provenance_eq_Procedure=proc)
 
+@sched_op([ExprCursorA(many=True)])
```
gilbo, line 718: Should this operation apply to a single location or many locations?

gilbo:
This scheduling op should probably have a more specific name than `commute`.  Maybe `commute_expr` is sufficient.  But are `+` and `*` commutativity the only possible thing that the phrase `commute_expr` might mean?

src/exo/API_scheduling.py
yamaguchi1024: I'm fine with both, do you have a suggestion?

yamaguchi1024:
I'm fine with both `commute` and `commute_expr`, latter is a bit harder to type though

gilbo (approved):

PR #225, opened 2022-08-03
Reproducer for blockcursor error

yamaguchi1024:
This test case fails with the following error:
```
E         File "<unknown>", line 1
E           y:_ ; for i in _:_
E                 ^
E       SyntaxError: invalid syntax
```

### DIFF
--- a/tests/test_schedules.py
+++ b/tests/test_schedules.py
@@ -7,6 +7,16 @@ from exo.libs.memories import GEMM_SCRATCH
 from exo import ParseFragmentError
 from exo.stdlib.scheduling import *
 
+def test_reorder_stmts():
+    @proc
+    def foo():
+        x : R[30]
+        y : R[30]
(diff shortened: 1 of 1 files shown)


gilbo:
This is an inherent limitation of the Python parser, which is being used to parse the string.  It doesn't have to do with `Exo` code.  It would be nice if we could do some limited pre-processing for patterns to allow for some of these short-hands, but that's more of a feature request than a bug report.

The most direct way to resolve this issue is to replace the string
```
"y : _ ; for i in _ : _"
```
with the string
```
("y : _\n"
 "for i in _ : _")
```

However, if this is for the purposes of calling `reorder_stmts` then you may be better off calling `p.find(pattern_for_one_stmt).expand(-1)` or `expand(1)` depending on which statement you want to select relative to the one you found.

If this resolves the issue in this PR, please close out the PR.

PR #230, opened 2022-09-26
Check generated C code from apps

alexreinking:
This was more complicated than initially anticipated. The Exo C backend could order its outputs non-deterministically.


yamaguchi1024 (approved):
LGTM, thanks!

PR #236, opened 2022-10-03
Implement merge writes

### LINKED ISSUE #205: Implement merge_reduce
Open to suggestions for a better name
```
loc = expr1
loc += expr2
--->
loc = expr1 + expr2
```

skeqiqevian:
Closes #205. Supports array accesses and scalar. Will not perform the merge if the LHS of the first statement is used in the RHS of the second statement.

rewrite:
        `a = b`
        `a = c`
            ->
        `a = c`
        ----------------------
        `a += b`
        `a = c`
            ->
        `a = c`
        ----------------------
        `a = b`
        `a += c`
            ->
        `a = b + c`
        ----------------------
        `a += b`
        `a += c`
            ->
        `a += b + c`


yamaguchi1024:
Looks good to me. @gilbo @alexreinking Do you want to comment?

alexreinking:
This is a useful change, but I think it would be a _lot_ more useful if it were generalized slightly. I think this should be something more like `fuse_writes` which takes two adjacent writes to the same location and condenses them down to a single write. There are four cases:

1. Write-Write. The second write wins, the first write is deleted.
2. Write-Accum. This PR.
3. Accum-Write. The write wins, the accum is deleted.
4. Accum-Accum. `x += y ; x += z` -> `x += y + z`.

Then we can implement a generalization to `n` consecutive writes on top of this.

yamaguchi1024:
@alexreinking Generalizing this more sounds useful/reasonable, @skeqiqevian can you take a look?

alexreinking (changes requested):
Will approve after appropriate generalization.

src/exo/LoopIR_scheduling.py
```diff
@@ -195,7 +195,41 @@ def map_e(self,e):
...
+
+            if len(stmt1.idx) != len(stmt2.idx):
+                raise SchedulingError("expected the LHS indices to be the same.")
+            try:
+                for i, j in zip(stmt1.idx, stmt2.idx):
+                    Check_ExprEqvInContext(proc, [stmt1, stmt2], i, j)
+            except SchedulingError:
+                raise SchedulingError("expected the LHS indices to be the same.")
+
+            self.new_rhs = LoopIR.BinOp("+", stmt1.rhs, stmt2.rhs,T.i32, stmt1.srcinfo)
```
yamaguchi1024, line 212: We should also check data dependencies on the rhs

alexreinking: And write tests for those cases, e.g.
```
a = b
a += a
```
does not rewrite to `a = a + b`, but rather `a = b + b`

src/exo/LoopIR_scheduling.py
```diff
@@ -194,8 +194,52 @@ def map_e(self,e):
...
+        assert False, "bad case"
+
+class _DoMergeWrites(LoopIR_Rewrite):
+    def __init__(self, proc, stmt1, stmt2):
+        try:
+            assert len(stmt1.idx) == len(stmt2.idx)
+            for i, j in zip(stmt1.idx, stmt2.idx):
+                Check_ExprEqvInContext(proc, [stmt1, stmt2], i, j)
+        except SchedulingError:
+            raise SchedulingError("expected the left hand side's indices to be the same.")
```
alexreinking, line 218: You are handling an error by raising an error. The right to do that is `raise ... from` to tell the Python runtime that there aren't _two_ errors.
```suggestion
        except SchedulingError as e:
            raise SchedulingError("expected the left hand side's indices to be the same.") from e
```

src/exo/LoopIR_scheduling.py
```diff
@@ -194,8 +194,52 @@ def map_e(self,e):
...
+        self.new_rhs = LoopIR.BinOp("+", stmt1.rhs, stmt2.rhs, stmt1.type, stmt1.srcinfo)
+        self.s1 = stmt1
+        self.s2 = stmt2
+
+        super().__init__(proc)
+
+        self.proc = InferEffects(self.proc).result()
 
+    def map_stmts(self, stmts):
+        if type(self.s2) is LoopIR.Assign:
```
alexreinking, line 233: Please use `isinstance` instead of type function checks.

src/exo/LoopIR_scheduling.py
```diff
@@ -194,8 +194,52 @@ def map_e(self,e):
...
+class _DoMergeWrites(LoopIR_Rewrite):
+    def __init__(self, proc, stmt1, stmt2):
+        try:
+            assert len(stmt1.idx) == len(stmt2.idx)
+            for i, j in zip(stmt1.idx, stmt2.idx):
+                Check_ExprEqvInContext(proc, [stmt1, stmt2], i, j)
+        except SchedulingError:
+            raise SchedulingError("expected the left hand side's indices to be the same.")
+
+        if any([stmt1.name == name and stmt1.type == typ for name, typ in get_reads(stmt2.rhs)]):
```
alexreinking, line 220: Drop the `[]` to avoid materializing a list here.

src/exo/LoopIR_scheduling.py
```diff
@@ -194,8 +194,52 @@ def map_e(self,e):
...
+        self.s1 = stmt1
+        self.s2 = stmt2
+
+        super().__init__(proc)
+
+        self.proc = InferEffects(self.proc).result()
 
+    def map_stmts(self, stmts):
+        if type(self.s2) is LoopIR.Assign:
+            for i in range(len(stmts)):
```
alexreinking, line 234: Use `for i, s in enumerate(stmts)` to avoid unnecessary indexing.

tests/golden/test_schedules/test_merge_writes_thing.txt
```diff
@@ -0,0 +1,5 @@
+def bar(x: R[5] @ DRAM, y: R[3] @ DRAM):
```
alexreinking, line 1: Please delete stale golden output.

alexreinking (changes requested):
Looks pretty good, there's just some weird Python details to be aware of and fix here.

alexreinking:
@skeqiqevian please push an empty commit to make CI run.
```
git commit --allow-empty -m "trigger CI"
```
To avoid this in the future, please run:
```
python -m pip install -r requirements.txt
pre-commit install
```
Then the code formatter will run whenever you commit.

alexreinking (approved):
Looks good. Ready to merge after CI runs

PR #237, opened 2022-10-03
Fixed bind_expr bug with LHS indices

skeqiqevian:
Originally, `bind_expr`'s index checking code vacuously returned True, so in the code below, we would replace both `w[i]` and `w[i+1]` with `w_tmp`.
```
@proc
    def bar(n: size, x: i8[n], y: i8[n], z: i8[n]):
        for i in seq(0, n-1):
            w : i8[n]
            w[i+1] = x[i] + y[i]
            w[i] = x[i] + y[i]
            z[i] = w[i]

    bar = bind_expr(bar, 'w[i]', 'w_tmp', cse=True)
```
With this fix, only the LHS of `w[i] = x[i] + y[i]` will be replaced.


alexreinking:
This change is incorrect.

Here's the "before" code:
```
w : i8[n]
w[i+1] = x[i] + y[i]
w[i] = x[i] + y[i]
z[i] = w[i]
```
Here's the "after" code:
```
w: i8[n] @ DRAM         # L1
w[i + 1] = x[i] + y[i]  # L2
w_tmp = x[i] + y[i]     # L3
w_tmp: R                # L4
w_tmp = w[i]            # L5
z[i] = w_tmp            # L6
```
Here are the problems that I see:
  1. `w_tmp` is written-to (L3) before it is allocated (L4)
  2. The value `x[i] + y[i]` is computed (L3) and then overwritten by the old value (L5).
  3. No write to `w[i]` takes place in the "after" code.
  4. `bind_expr` isn't supposed to touch the RHS of expressions, anyway.

alexreinking (changes requested):
See my comment on the main PR. The test here shows invalid code.

alexreinking (approved):
The bugs I pointed out are real, but I don't think this patch causes them. If anything, it's improving a bad situation. However, this feature needs to be fixed, and tests need to be written to avoid regressions. We can do this in another PR.

See issue: https://github.com/exo-lang/exo/issues/238

PR #240, opened 2022-10-04
Codegen cleanup

### LINKED ISSUE #95: Empty structs differ in C/C++
Empty structs, such as:

```
typedef struct sgemm_Context {
} sgemm_Context;
```

have different sizes in C and C++. This could lead to ABI issues down the line. If we would generate an empty struct, we should either:

1. use `void*` instead, or
2. add an `int_fast32_t dummy;` field

alexreinking:
### Add `const` to function signatures

When a function does not write or accumulate to a parameter it's passed, it should tell the C compiler about this by declaring it `const`. In some cases that can promote inlining or better codegen and definitely improves the ergonomics of _using_ exo functions by allowing callers to declare their local variables const in the first place. It also plays nicer with C's type system generally.

### Use `void` instead of an empty struct

This prevents an actual ABI issue whereby a C compiler is allowed to give such a struct a size of 0, but C++ must give it a size of 1.

Fixes #95 

### Only generate `_floor_div` and `_clamp_32to8` when needed.

Self-explanatory. Also uses `exo_` prefixes since leading underscores are generally sketchy in C.

### DIFF
--- a/tests/golden/pldi22/test_gemmini_conv_ae/test_conv_ae.txt
+++ b/tests/golden/pldi22/test_gemmini_conv_ae/test_conv_ae.txt
@@ -41,17 +41,21 @@ typedef struct c_code_str_Context {
     } ConfigStore;
 
 } c_code_str_Context;
-struct exo_win_2i32{
-    int32_t *data;
-    int_fast32_t strides[2];
+struct exo_win_2i32c{
+    const int32_t * const data;
+    const int_fast32_t strides[2];
 };
 struct exo_win_2i8{
-    int8_t *data;
-    int_fast32_t strides[2];
+    int8_t * const data;
+    const int_fast32_t strides[2];
 };
-struct exo_win_3i8{
-    int8_t *data;
-    int_fast32_t strides[3];
+struct exo_win_2i8c{
+    const int8_t * const data;
+    const int_fast32_t strides[2];
+};
+struct exo_win_3i8c{
+    const int8_t * const data;
+    const int_fast32_t strides[3];
 };
 // conv_on_gemmini(
 //     output : i8[4,56,56,64]  @DRAM,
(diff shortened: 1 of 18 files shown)


alexreinking:
@yamaguchi1024 - I'm requesting your review just so this is in your queue, but it's not ready to land yet.

tests/golden/pldi22/test_gemmini_conv_ae/test_conv_ae.txt
```diff
@@ -396,7 +392,7 @@ gemmini_extended_compute_preloaded((uint32_t)(&*(int8_t*)((uint64_t)( ((uint32_t
             }
           }
           if (orow_ii + 7 * orow_io == 0 || false) {
-            gemmini_extended4_config_ld(((struct exo_win_2i8){ (int8_t*)&inp[(b) * (58 * 58 * 64) + (1 + orow_ii + 7 * orow_io + 28 * orow_o) * (58 * 64) + (2 + 16 * ocol_o) * (64) + (0) * (1)], { 64, 1 } }).strides[0]*1, 1.0f, 0, (16), 2);
+            gemmini_extended4_config_ld(((struct exo_win_2i8c){ &inp[(b) * (58 * 58 * 64) + (1 + orow_ii + 7 * orow_io + 28 * orow_o) * (58 * 64) + (2 + 16 * ocol_o) * (64) + (0) * (1)], { 64, 1 } }).strides[0]*1, 1.0f, 0, (16), 2);
```
yamaguchi1024, line 395: What does `c` stands for?

yamaguchi1024:
Great changes!!
Were we able to remove some of the `-fno-` 's in the cmake scripts for Gemmini as a result?

tests/golden/pldi22/test_gemmini_conv_ae/test_conv_ae.txt
alexreinking: Const

alexreinking:
Not yet. That will be my next PR or two

yamaguchi1024 (approved):
LGTM, thanks!

PR #243, opened 2022-10-05
Run Gemmini tests in CI

alexreinking:
Recreating this PR because GHA is mysteriously not running on the original (#239)


alexreinking:
@yamaguchi1024 - this is ready. We still need to improve the workflow to fix warnings, skip the `#define` hacks, etc, but at least we are _running_ them now.

yamaguchi1024 (approved):
LGTM

PR #244, opened 2022-10-06
Rework Gemmini testing into standalone app

alexreinking:
This _significantly_ reduces the amount we care about what's in gemmini-rocc-tests. If you don't count the two new golden outputs, this PR actually _removes_ 1,086 lines of code from the repository and also dramatically shrinks the patch in our gemmini-rocc-tests fork.

Assuming you have the environment variable `RISCV` set appropriately, to build and run the Gemmini tests (bare-metal), all you need to do is:

```
$ cmake -S apps/gemmini -B build -DCMAKE_BUILD_TYPE=Release
$ cmake --build build
$ cd build && ctest
```

The build takes care of downloading our fork of gemmini-rocc-tests. It uses header paths to make sure the tests get our copies of functions. It improves on the upstream build system by not compiling the same sources many times.

Still worth noting: the gemmini schedules take a _long_ time to execute. We need some sort of parallelism here.


yamaguchi1024:
@alexreinking Why did most of the files under gemmini deleted? I removed rotten tests in the previous PR, and the remaining tests should not be removed.

alexreinking:
> Why did most of the files under gemmini deleted? I removed rotten tests in the previous PR, and the remaining tests should not be removed.

They weren't _removed_, they were moved into a standalone app.

alexreinking:
Oh, actually, I see I was overzealous in dropping the AE and paper tests... those should probably just become golden-output tests. There's enough running in Spike now between the gemmini-rocc-tests and the individual conv/matmul tests.

alexreinking:
I restored the PLDI paper/ae tests underneath `tests/pldi22`. I noticed that test_gemmini_matmul_paper.py doesn't compile via `c_code_str` because of a memory type mismatch... so that test is just golden for the Exo scheduled output, rather than the C code like the others.

yamaguchi1024 (approved):

PR #254, opened 2022-10-21
Bind expr fix

### LINKED ISSUE #238: bind_expr with CSE is broken
Per the discussion in https://github.com/exo-lang/exo/pull/237 the CSE option for bind_expr incorrectly factors out right-hand-sides and generally breaks well-formedness of the code. This needs to be fixed.

skeqiqevian:
CSE no longer replaces the LHS, and will only replace the RHS until it encounters a read to any buffer that the expression depends on.

Fixes #238


src/exo/LoopIR_scheduling.py
```diff
@@ -1169,6 +1169,22 @@ def map_e(self, e):
             return super().map_e(e)
 
 
+def get_reads(e):
```
alexreinking, line 1172: Is this redundant with the other PR, #236?

src/exo/LoopIR_scheduling.py
```diff
@@ -1257,28 +1279,14 @@ def map_s(self, s):
...
-                            cast=None,
-                            idx=[],
-                            rhs=self.apply_e(s.rhs),
-                            eff=None,
-                        )
-                    ]
-                except SchedulingError:
-                    pass
+            if (new_rhs := self.map_e(s.rhs)) is None:
+                new_rhs = s.rhs
```
alexreinking, line 1283: Better would be to return `None` when `new_rhs` is `None` because otherwise the `return [s.update(rhs=new_rhs)]` makes a pointless copy of `s`. Naturally, don't skip the `self.found_write` logic when making this change.

alexreinking (changes requested):
Another small fix.

src/exo/LoopIR_scheduling.py
skeqiqevian: Yeah it's the exact same function

yamaguchi1024 (approved):

PR #255, opened 2022-10-24
Deprecate PatternParser

### LINKED ISSUE #249: Rework parser to have the same path for proc and fragment
ParseFragment should parse window as well

yamaguchi1024:
This PR unifies PatternParser to Parser class 

Fixes #249

### DIFF
--- a/src/exo/pyparser.py
+++ b/src/exo/pyparser.py
@@ -136,7 +136,7 @@ class Parser:
         getsrcinfo,
         is_fragment=False,
         func_globals=None,
-        srclocals=None,
+        srclocals=ChainMap(),
         as_func=False,
         as_config=False,
         instr=None,
@@ -150,7 +150,10 @@ class Parser:
 
         self.push()
 
-        if not is_fragment:
+        if is_fragment:
+            self.AST = PAST
+        else:
+            self.AST = UAST
             # add builtins
             self.locals["sin"] = sin
             self.locals["relu"] = relu
@@ -179,12 +182,10 @@ class Parser:
         return self._cached_result
 
     def push(self):
-        if not self.is_fragment:
-            self.locals = self.locals.new_child()
+        self.locals = self.locals.new_child()
 
     def pop(self):
-        if not self.is_fragment:
-            self.locals = self.locals.parents
+        self.locals = self.locals.parents
 
     # - # - # - # - # - # - # - # - # - # - # - # - # - # - # - #
     # parser helper routines
@@ -665,10 +666,7 @@ class Parser:
                 orelse = self.parse_stmt_block(s.orelse)
                 self.pop()
 
-                if self.is_fragment:
-                    rstmts.append(PAST.If(cond, body, orelse, self.getsrcinfo(s)))
-                else:
-                    rstmts.append(UAST.If(cond, body, orelse, self.getsrcinfo(s)))
+                rstmts.append(self.AST.If(cond, body, orelse, self.getsrcinfo(s)))
 
             # ----- Sub-routine call parsing
             elif (
@@ -732,10 +730,7 @@ class Parser:
 
             # ----- Pass no-op parsing
             elif isinstance(s, pyast.Pass):
-                if self.is_fragment:
-                    rstmts.append(PAST.Pass(self.getsrcinfo(s)))
-                else:
-                    rstmts.append(UAST.Pass(self.getsrcinfo(s)))
+                rstmts.append(self.AST.Pass(self.getsrcinfo(s)))
 
             # ----- Stmt Hole parsing
             elif (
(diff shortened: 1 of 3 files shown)


src/exo/pyparser.py
```diff
@@ -168,10 +179,12 @@ def result(self):
...
 
     def push(self):
-        self.locals = self.locals.new_child()
+        if not self.is_fragment:
+            self.locals = self.locals.new_child()
 
     def pop(self):
-        self.locals = self.locals.parents
+        if not self.is_fragment:
+            self.locals = self.locals.parents
```
alexreinking, line 187: Why doesn't this work when `is_fragment` is true?

src/exo/pyparser.py
```diff
@@ -1272,10 +959,14 @@ def parse_expr(self, e):
...
                 assert False, "unrecognized op"
+
             if op not in front_ops:
                 self.err(e, f"unsupported binary operator: {op}")
 
-            return PAST.BinOp(op, lhs, rhs, self.getsrcinfo(e))
+            if self.is_fragment:
+                return PAST.BinOp(op, lhs, rhs, self.getsrcinfo(e))
+            else:
+                return UAST.BinOp(op, lhs, rhs, self.getsrcinfo(e))
```
alexreinking, line 970: A better design for this would be to write something like:
```
if is_fragment:
    self.AST = PAST
else:
    self.AST = UAST
```
and then just using (e.g.) `self.AST.BinOp` everywhere

src/exo/pyparser.py
```diff
@@ -525,78 +557,102 @@ def parse_stmt_block(self, stmts):
...
                     self.err(s.target, "expected simple name for iterator variable")
-                itr = Sym(s.target.id)
-                self.locals[s.target.id] = itr
-                hi = self.parse_loop_cond(s.iter)
+
+                if self.is_fragment:
+                    itr = s.target.id
+                else:
+                    itr = Sym(s.target.id)
+                    self.locals[s.target.id] = itr
```
alexreinking, line 645: Is this the same as `self.locals[s.target.id] = s.target.id.copy()`?

alexreinking (changes requested):
This is great! Just for the sake of maintenance, I'd like the `self.AST` change applied throughout.

src/exo/pyparser.py
yamaguchi1024: Because there is no local variables when parsing fragment

src/exo/pyparser.py
yamaguchi1024: Yup!

src/exo/pyparser.py
yamaguchi1024: Ahh sounds good!

src/exo/pyparser.py
yamaguchi1024: Oh actually no, we need to create a symbol from string here

src/exo/pyparser.py
alexreinking: I think it would be better to initialize `self.locals = srclocals or ChainMap()` in the constructor. We might want to allow matching patterns like `for i in _: x[i] = _` down the line. The push and pop functions are used to control the scopes of for loop variables, too. Having these functions sometimes do nothing will be confusing.

I think it's better to solve this problem with data (init with empty map) than with control (check which AST we're parsing).

alexreinking: Hah! You beat me to it!

src/exo/pyparser.py
alexreinking: Ok, cool. I wasn't sure... that class has too many similar-but-different ways to work with it.

src/exo/pyparser.py
```diff
@@ -133,9 +133,10 @@ class Parser:
     def __init__(
         self,
         module_ast,
-        func_globals,
-        srclocals,
         getsrcinfo,
+        is_fragment=False,
+        func_globals=None,
+        srclocals=ChainMap(),
```
alexreinking, line 139: This isn't safe because `ChainMap` is mutable... this should be `None` and then you should have something like `srclocals = srclocals or ChainMap()` in the constructor body.

The issue is that the values of default arguments are stored in the _function_ object and so are shared across all calls. If something gets added to the base chain map in one call, it will be seen by all subsequent calls.

src/exo/pyparser.py
```diff
@@ -525,78 +558,102 @@ def parse_stmt_block(self, stmts):
...
+
+                cond = self.parse_loop_cond(s.iter)
                 body = self.parse_stmt_block(s.body)
-                self.pop()
 
-                rstmts.append(UAST.ForAll(itr, hi, body, self.getsrcinfo(s)))
+                if self.is_fragment:
+                    rstmts.append(PAST.Seq(itr, cond, body, self.getsrcinfo(s)))
+                else:
+                    rstmts.append(UAST.ForAll(itr, cond, body, self.getsrcinfo(s)))
```
yamaguchi1024, line 654: We cannot get rid of this for now because UAST is still using ForAll, we should probably deprecate that too

src/exo/pyparser.py
```diff
@@ -145,21 +146,34 @@ def __init__(
         self.globals = func_globals
         self.locals = srclocals
```
alexreinking, line 147: This can be `self.locals = srclocals or ChainMap()`

alexreinking (changes requested):
One last little Python "gotcha".

src/exo/pyparser.py
yamaguchi1024: Ah ok that makes sense!

src/exo/pyparser.py
yamaguchi1024: Yup will do!

alexreinking (approved):
Awesome!

PR #258, opened 2022-10-24
Deprecate UAST.ForAll

### LINKED ISSUE #257: Deprecate ForAll from UAST
Other AST types use `Seq` and this is confusing.

yamaguchi1024:
Fixes #257
This was simpler than I thought


alexreinking (approved):
Great! I didn't think this would be too bad.

PR #259, opened 2022-10-24
Fix lift_alloc to emit error when n_lifts is too large

### LINKED ISSUE #247: Fix lift_alloc to emit error when n_lifts is too large

yamaguchi1024:
Fixes #247

### DIFF
--- a/src/exo/LoopIR_scheduling.py
+++ b/src/exo/LoopIR_scheduling.py
@@ -1807,7 +1807,8 @@ class _DoLiftAllocSimple(LoopIR_Rewrite):
 # --------------------------------------------------------------------------- #
 # Lift Allocation scheduling directive
 
-
+# TODO: Implement autolift_alloc's logic using high-level scheduling metaprogramming and
+#       delete this code
 class _LiftAlloc(LoopIR_Rewrite):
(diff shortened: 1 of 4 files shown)


src/exo/LoopIR_scheduling.py
```diff
@@ -1848,12 +1849,16 @@ def idx_mode(self, access, orig):
...
     def map_s(self, s):
         if s is self.alloc_stmt:
-            # mark the point we want to lift this alloc-stmt to
-            n_up = min(self.n_lifts, len(self.ctrl_ctxt))
-            self.lift_site = self.ctrl_ctxt[-n_up]
+            if self.n_lifts > len(self.ctrl_ctxt):
+                raise SchedulingError(
+                    "specified lift level {self.n_lifts} "
+                    + "is higher than the number of loop "
+                    + "{len(self.ctrl_ctxt)}"
```
alexreinking, line 1856: Missing an `f` here for format string. Also `+` is not needed.

alexreinking (changes requested):
Tiny bug in error message.

src/exo/LoopIR_scheduling.py
yamaguchi1024: Woops thanks!

alexreinking (approved):

PR #260, opened 2022-10-25
Expose cursor to scheduling; Delete four rotten primitives

yamaguchi1024:


alexreinking (approved):
Assuming CI likes this, looks good. I'm a bit surprised add/merge guard weren't used.

PR #266, opened 2022-11-01
Change scheduling API to take stmt cursors

yamaguchi1024:


alexreinking (approved):
Looks good for now. I still think we're just going to have to [draw the whole owl](https://i.kym-cdn.com/photos/images/newsfeed/000/572/078/d6d.jpg) for some of these scheduling directives just to see what they look like,

PR #267, opened 2022-11-02
Add can_read to memories

yamaguchi1024:
I have no idea why we didn't discover this until now...

### DIFF
--- a/tests/test_neon.py
+++ b/tests/test_neon.py
@@ -14,15 +14,14 @@ from exo.memory import MemGenError
 import numpy as np
 
 
-@pytest.mark.isa("neon")
-def test_neon_can_read(compiler):
+def test_neon_can_read():
     @proc
     def read_neon(n: size, dst: R[n] @ DRAM, src: R[n] @ Neon4f):
         for i in seq(0, n):
             dst[i] = src[i]
 
     with pytest.raises(MemGenError, match="cannot read"):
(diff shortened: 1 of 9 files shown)


yamaguchi1024:
FYI this is coming from Adrian's email

alexreinking:
I think the problem is with this line:

https://github.com/exo-lang/exo/blob/6e7419ee9642ed92993a3403b7bde7e85b24c701/src/exo/memory.py#L95

Should be `return`, not `raise`... what issue was Adrian seeing? I'm not sure I got the email.

yamaguchi1024:
@alexreinking The line is fine, it's trying to say that whoever inherits this class should specify if the memory is readable or not, which I think is fair.

tests/test_neon.py
```diff
@@ -9,10 +9,22 @@
 from exo import proc
 from exo.platforms.neon import *
 from exo.stdlib.scheduling import *
+from exo.memory import MemGenError
 
 import numpy as np
 
 
+@pytest.mark.isa("neon")
```
alexreinking, line 17: This test doesn't actually run anything on NEON, it just needs to get as far as generating C. Therefore it should never be skipped and you can check that `read_neon.c_code_str()` throws the expected `MemGenError`.

alexreinking (changes requested):
No reason to skip this test.

tests/test_neon.py
yamaguchi1024: That makes sense!

alexreinking (approved):
Looks good when green

PR #276, opened 2022-11-04
Fix inline_window to correctly handle stride dimension

yamaguchi1024:
This was the source problem of  gemmini stride issue in `api-audit-and-fix` we saw earilier

### DIFF
--- a/src/exo/LoopIR_scheduling.py
+++ b/src/exo/LoopIR_scheduling.py
@@ -1074,6 +1074,19 @@ class _InlineWindow(Cursor_Rewrite):
 
         return new_idxs
 
+    def calc_dim(self, dim):
+        assert dim < len(
+            [w for w in self.win_stmt.rhs.idx if isinstance(w, LoopIR.Interval)]
+        )
+
+        new_dim = 0
+        for w in self.win_stmt.rhs.idx:
+            if isinstance(w, LoopIR.Interval):
+                dim -= 1
+            if dim == -1:
+                return new_dim
+            new_dim += 1
+
     def map_s(self, sc):
         s = sc._node()
         if s is self.win_stmt:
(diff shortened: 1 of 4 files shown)


src/exo/LoopIR_scheduling.py
```diff
@@ -1074,6 +1074,19 @@ def calc_idx(self, idxs):
...
+            [w for w in self.win_stmt.rhs.idx if isinstance(w, LoopIR.Interval)]
+        )
+
+        new_dim = 0
+        for w in self.win_stmt.rhs.idx:
+            if isinstance(w, LoopIR.Interval):
+                dim -= 1
+            if dim == -1:
+                return new_dim
+            new_dim += 1
```
alexreinking, line 1088: This is really quite confusing... it would benefit a lot from a comment describing how it works and a specifically targeted test.

src/exo/platforms/gemmini.py
```diff
@@ -544,15 +544,15 @@ def make_ld_i8_block(name, ld_id, p=ld_i8_block, stride_val=None):
     "ld_i8_block_id1_v2", 1, stride_val="stride(src,0)"
 )
 ld_i8_block_id1_s2_v2 = make_ld_i8_block(
-    "ld_i8_block_id1_s2_v2", 1, stride_val="stride(src,2)"
+    "ld_i8_block_id1_s2_v2", 1, stride_val="stride(src,0)"
```
alexreinking, line 547: Do these really not make it into the golden outputs of _any_ test?

alexreinking (changes requested):

alexreinking:
Also, yay, I fixed the code-coverage bot!

src/exo/platforms/gemmini.py
yamaguchi1024: Yeah!

alexreinking (approved):
Awesome, thanks!

PR #280, opened 2022-11-07
Mark slow tests as such in PyTest.

alexreinking:
This is just a developer convenience.

Run pytest with `pytest -m "not slow"` to run only the fast tests. With this, the whole test suite takes about a minute on my laptop. Works well with `-n $(nproc)` for parallelism. Does NOT work well with `--update-golden` because all golden outputs should typically be updated at once.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,9 @@ filterwarnings = [
 ]
 norecursedirs = ['dependencies']
 pythonpath = "src/"
+markers = [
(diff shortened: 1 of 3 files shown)


yamaguchi1024 (approved):
@alexreinking This looks great! This still runs Gemmini tests by default, right?

alexreinking:
@yamaguchi1024 - yes, it's all opt-in and doesn't change how anything currently works. Does not affect CI at all. This just adds a label to the tests that take more than 30 seconds on my machine so that they can be manually excluded.

PR #288, opened 2022-11-12
Fix const-window argument passing

alexreinking:
Discovered at the Hackathon yesterday. We don't have enough tests that stress parameter passing, which is why this wasn't caught in the first place...

### DIFF
--- a/tests/golden/test_apps/test_gemmini_matmul.txt
+++ b/tests/golden/test_apps/test_gemmini_matmul.txt
@@ -54,6 +54,10 @@ typedef struct test_case_Context {
     } ConfigStore;
 
 } test_case_Context;
+struct exo_win_2i32{
+    int32_t * const data;
+    const int_fast32_t strides[2];
+};
 struct exo_win_2i32c{
     const int32_t * const data;
     const int_fast32_t strides[2];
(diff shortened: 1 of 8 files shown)


tests/golden/test_apps/test_gemmini_matmul.txt
```diff
@@ -66,8 +70,8 @@ struct exo_win_2i8c{
     const int8_t * const data;
     const int_fast32_t strides[2];
 };
-struct exo_win_3i8c{
-    const int8_t * const data;
+struct exo_win_3i8{
```
alexreinking, line 73: In this test and the others, these structs aren't actually _used_ anywhere. I'm guessing they come from instr signatures, but I'm not totally sure.

yamaguchi1024: Just curious, was there any warning or performance consequence from not having const?

alexreinking: Nothing we rigorously measured. But there are still a great many reasons for _having_ const.

1. Compilers _can indeed_ optimize based on const-ness
2. Having `const` enables the C compiler to check our generated code. If we misplace a const the C compiler will warn or error, and that means our analysis somewhere is wrong. This helps us catch bugs.
3. Having accurate `const` information makes it easier to integrate into existing C codebases. For instance, someone might have a const-qualified buffer in their program. They would have to cast away that const to call an Exo function. That's a recipe for unexpected / undefined behavior.

and so on...

yamaguchi1024 (approved):
LGTM, thanks!

PR #292, opened 2022-11-16
Improve exocc command line interface

### LINKED ISSUE #264: exocc overwrites all the prints when there is an error
`exocc` should flush the print along the way

### LINKED ISSUE #270: exocc: An option to disable C compilation
I'd like to see something like
```
exocc test.py
```
only runs the scheduling frontend, and
```
exocc test.py -o test.c
```
runs the C compilation as well, but exact design is open to discussion

alexreinking:
Adds two output modes:

  1. -o/--outdir OUTDIR [-s/--stem STEM]
  2. -oc/--cfile CFILE -oh/--hfile HFILE

In the first mode, the user specifies at least an OUTDIR for the .c and .h files. If they are compiling only one Exo source file, then the STEM is taken from that file. Otherwise, STEM must be provided.

In the second mode, the user specifies exact paths for the .c and .h files as CFILE and HFILE, respectively. This is to support build systems (like Bazel and Meson) whose rules for relating input file names to output file names for code generators are more stringent.

Additionally, adds --debug DEBUG option where a DEBUG mode may be specified. For now, only `schedule` is understood as a debug mode. When specified, the top-level funcs are printed in their final, scheduled form. In a follow-up PR, this will be extended to the transitive closure of all procs that will be compiled.

Fixes #264
Fixes #270


yamaguchi1024:
Looks good, thanks! Can you document this exocc interface change to README?

yamaguchi1024:
We should come back to this PR at some points, seems useful to merge

yamaguchi1024:
Closing stale PR. Some of this PR was merged with #659

PR #191, opened 2022-05-11
Api Audit and Fixing Up Effect Checks

gilbo:
This PR will collect all of my work to make sure the new effect checks are working.

This PR will not yet try to deprecate or strip out the old effect system.

This PR will also not yet try to replace the old bounds check with the new effect-checking machinery.  (Another PR should replace the bounds checking / assert checking, and then yet another PR can strip out the old effect checking stuff entirely)

PR #293, opened 2022-11-20
Effects merge attempt 2

yamaguchi1024:
This PR only includes the existing effect change from #191, but it still slows our testing by 7.6x. 
What should we do?


alexreinking:
Merging with #294 open

alexreinking (approved):

PR #297, opened 2023-01-04
Half precision support added

adcastel:


yamaguchi1024 (changes requested):
Thanks Adrian! Can you add tests for f16 as well?

src/exo/LoopIR.py
```diff
@@ -418,6 +427,8 @@ def shape(t):
 def ctype(t):
     if isinstance(t, T.Num):
         assert False, "Don't ask for ctype of Num"
+    elif isinstance(t, T.F16):
+        return "__fp16"
```
gilbo, line 431: It is unclear whether or not this induces the intended semantics of half-precision floating point, and it is unclear whether or not it is portable across C compilers. (see also latter occurrence of this same pattern in `configs.py`; note for all: the reference point for this translation to C types should be deduplicated in the code base)

See [https://gcc.gnu.org/onlinedocs/gcc-4.9.4/gcc/Half-Precision.html#:~:text=The%20__fp16%20type%20is,parameters%20of%20type%20__fp16%20.](https://gcc.gnu.org/onlinedocs/gcc-4.9.4/gcc/Half-Precision.html#:~:text=The%20__fp16%20type%20is,parameters%20of%20type%20__fp16%20.) for instance.

Clang supports something similar to that GCC feature.  However further note the following:
* GCC requires an additional compiler flag.  Therefore, switching to using `f16` in Exo code may cause downstream compilation to break.  This seems bad for a core compiler feature.  By contrast `f32` and `f64` translate (mostly without issue) to `float` and `double` (If you use the old `x87` floating point unit on some processors `f64` is actually allowed to compute on 80 bit registers and only store to 64, but thankfully this is no longer the case on modern machines.)
* "The `__fp16` type is a storage format only."  This means that you won't get consistent results between scalar code using `__fp16` and the Neon vectors using half-precision.

We have chatted some about abstracting out precision from the compiler.  Maybe we should schedule a working session (at least @yamaguchi1024 and I, but @adcastel and @alexreinking are also welcome) where we sort out a quick design for precision extensibility in Exo.  To give a quick sketch:
* Precisions can be specified by objects/classes encapsulating a set of methods, similar to memories
* Those methods can specify fragments of C code to be interpolated.
* Those methods can specify a particular datatype to be used (possibly a custom struct even)
* Those methods can specify how the basic arithmetic in Exo should be compiled/lowered.  This can include invoking whole functions intended to produce bitwise accurate simulations of e.g. `bf16` even if the underlying CPU hardware does not support that precision type directly.
* Open Question: How should the interaction between built-in functions and precision types be handled?
* Open Question: How should precision conversions between precision types be specified?  (Note there is also an Exo language design question about how Exo programmers should express casting/coercion, but we can postpone and separate that question from implementing "externalized precisions")

(observe that the last two open questions involve potential pairwise interactions between precisions and/or builtins.  That could generate a lot of code to maintain, so it's worth thinking a little harder to see if there's a more elegant solution than just specifying all pairwise behavior.)

alexreinking: >and it is unclear whether or not it is portable across C compilers

See the newest version of the GCC docs, here: https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/Half-Precision.html

> It is recommended that portable code use the `_Float16` type defined by ISO/IEC TS 18661-3:2015.

So this is _somewhat_ more standardized. It's not in C11, but this should be more portable than `__fp16`, which might be enough to get something working _now_.

>  To give a quick sketch: ...

I agree with this design.

> Open Question: How should the interaction between built-in functions and precision types be handled?

Exo is a meta-programming framework for (a subset of) C... so I think it's fine to bless the datatypes that are provided by the C standard (like `f32` and `f64`) with special features.

> Open Question: How should precision conversions between precision types be specified?

I think conversions should be handled via the instruction mechanism and should never implicitly coerce to each other. This might involve adding "expression" instructions rather than just statement instructions. I'm not too worried about pairwise interactions because only some of the points in that space are useful and/or the user will find it suffices to convert via `double` rather than directly (which while lossy might not be _too_ lossy).

adcastel: I used __fp16 because it worked on ARM hardware. If I use _Float16 with GCC 10 it doesn't work and, with float16_t name (also ARM nomenclature) you need to add the #include <arm_neon.h> to the .h file generated by exo.

I'm not a compiler expert so I cannot help with these decisions. But, as a user, I think that the architecture binding (eg. avx instructions of intel or intrinsic of ARM translation) should be part of the compiler and not with the replace method.
E.g. if I want to vectorize the instruction in loop I, the user should only say vectorize(for i in _: _) instead of p = replace(p, 'for i in _: _ ', neon_vld_4xf32) so the intrinsic is automatically selected by the exo runtime.

yamaguchi1024 (approved):
LGTM, thank you! Will merge when the CI passes

yamaguchi1024 (approved):
LGTM, thank you! Will merge when the CI passes

yamaguchi1024:
BLAS CI is failing because this PR is old and doesn't have a BLAS token. Merging it anyways

Issue #160, opened 2022-02-16
Get rid of stage_assn, double_fission, and data_reuse and replace it with stage_memory


yamaguchi1024:

yamaguchi1024:
bind_expr

PR #301, opened 2023-01-07
Replace bind expr with stage mem

SamirDroubi:
An attempt at addressing #160 

It removes all usages of `bind_expr` and replaces it with `stage_mem` or a new stdlib function `stage_expr`. Only remaining usages are in examples/demo.

`bind_expr` is used throughout the codebase to perform manual staging of expressions. Many of those staging patterns can be refactored to completely use `stage_mem`. However, there are a few situations where that's not possible.
1. Staging to do a broadcast (for example when you want to stage `A` in matrix multiply, see test_x86.py::avx2_sgemm_6x16).
3. Staging a binop expressions (see test_x86.py::simple_math_avx2_sched)
4. Staging the temporary buffer into a buffer with higher dimensions than the original buffer (see test_im2col.py::test_im2col)

1 and 2 code patterns could be refactored into one function which I added into stdlib called `stage_expr`. However, the function ended up being too parameterized and has a problem where it could accept multiple expression cursors in the first step of `bind_expr`, but the function is really assuming one expression cursor. 

For 3, I simply replaced the call to `bind_expr` with a call to `stage_mem` but left the remainder of the staging pattern.

It seems that it is not possible to completely deprecate `bind_expr` because it is necessary for usages of binop expressions staging as in 2. Moreover, it seems that the implementation of `stage_expr` is not a good idea at least in it is the current way it is written. Finally, for 3 it seems more natural to use `bind_expr` instead of `stage_mem` since in that case, `stage_mem` is behaving exactly like `bind_expr` by removing the store subsequent to the block. For cases like 1, we could try to remove the staging pattern with `stage_mem` and try to use other scheduling primitives to achieve broadcasting, but that seems unnecessary if we have to keep `bind_expr` anyways.

In short, it might be best to keep `bind_expr`, and only replace its usage with `stage_mem` where the entire staging pattern could be removed.

### DIFF
--- a/apps/aarch64/sgemm/sgemm.py
+++ b/apps/aarch64/sgemm/sgemm.py
@@ -104,10 +104,7 @@ neon_microkernel = stage_C_microkernel()
 
 def stage_A_B_microkernel(p=neon_microkernel):
     for buf in ("A", "B"):
-        p = bind_expr(p, f"{buf}[_]", f"{buf}_vec")
-        p = expand_dim(p, f"{buf}_vec", 4, "ji", unsafe_disable_checks=True)
-        p = lift_alloc(p, f"{buf}_vec")
-        p = fission(p, p.find(f"{buf}_vec[_] = _").after())
+        p = stage_expr(p, f"{buf}[_]", 4, "ji", f"{buf}_vec")
         p = set_memory(p, f"{buf}_vec", Neon4f)
     #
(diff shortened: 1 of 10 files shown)


SamirDroubi:
`stage_expr` doesn't seem like a good addition to stdlib. It has some problems as outlined above. `bind_expr` is useful to keep around.

PR #302, opened 2023-01-11
Amx memory

skeqiqevian:
This PR creates a `StaticMemory` class which implements a free list allocator so that for static memory such as AMX Tiles, the Exo compiler can perform "register allocation" during compilation. This `StaticMemory` implementation is a prototype, as it will likely change when we try to support Gemmini too.

I also implemented some memory safeguards:
 - `StaticMemory` cannot allocate more memory than available
 - `StaticMemory` allocation will fail if the current proc is not a leaf node (e.g. it has no sub-procs other than `instr`s). This is a stricter condition than what we really need, which is that no nested procs can both allocate static memory. 
 - `AMX_TILE` cannot allocate tiles larger than 16x64 bytes
 
and wrote corresponding tests for those.


src/exo/memory.py
```diff
@@ -145,3 +145,37 @@ def write(cls, s, lhs, rhs):
...
+    """
+    A free list implementation of register allocation for static
+    memory.
+    """
+
+    is_chunk_allocated = []
+
+    @classmethod
+    def init_state(cls, size):
+        cls.is_chunk_allocated = [False for i in range(size)]
```
alexreinking, line 163:
```suggestion
        cls.is_chunk_allocated = [False] * size
```

yamaguchi1024 (approved):

alexreinking (approved):

PR #304, opened 2023-01-13
Add Formatted Expressions

SamirDroubi:
Add support for allowing users to supply Formatted Expressions. Users can now supply expressions with holes in them along with a list of ExprCursors to use to fill the holes.

### DIFF
--- a/tests/test_schedules.py
+++ b/tests/test_schedules.py
@@ -1544,15 +1544,22 @@ def test_formatted_expr_3(golden):
             tmp = 1.0
             arr[i] = tmp
 
-    with pytest.raises(ParseFragmentError, match="Too many holes in expression"):
+    with pytest.raises(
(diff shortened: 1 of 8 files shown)


tests/test_schedules.py
```diff
@@ -1484,3 +1484,68 @@ def accum(out: R[4, 16, 16], w: R[16], im: R[16]):
...
+
+def test_formatted_expr_3(golden):
+    @proc
+    def bar(n: size, arr: R[n] @ DRAM):
+        for i in seq(0, n):
+            tmp: R
+            tmp = 1.0
+            arr[i] = tmp
+
+    with pytest.raises(ParseFragmentError, match="Too many holes in expression"):
```
yamaguchi1024, line 1534: This error message is a bit ambiguous, I think it should be "The number of holes does not match with expressions"

yamaguchi1024 (approved):

PR #309, opened 2023-01-30
Implement internal node-cursor forwarding

alexreinking:
Implements node-only forwarding for internal cursors over the following edit operations:

* Insert
* Delete
* Replace
* Move
* Wrap

The implementation of move is _complicated_. Unexpectedly so. I've tried to write useful tests for all of these, but we might still encounter bugs. We need to stay diligent about adding test cases for bugs we fix.

Closes #278

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,9 +10,9 @@ jobs:
         runner: [ ubuntu-latest, macos-latest ]
         include:
           - runner: ubuntu-latest
-            CC: clang-15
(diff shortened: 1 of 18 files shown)


yamaguchi1024 (approved):
LGTM if it's ready to merge. Can #278 be closed because of this?

alexreinking:
> LGTM if it's ready to merge. Can #278 be closed because of this?

Yes. This is a better implementation.

Issue #312, opened 2023-01-31
Simplify should simplify mod expression


yamaguchi1024:
```
Ctile[(ii + 16 * ii_unroll) / 16,
(ji + 16 * ji_unroll) / 16,
(ii + 16 * ii_unroll) % 16,
(ji + 16 * ji_unroll) %
16] += a * b
```

`(ii + 16 * ii_unroll) % 16` should be simplified to `ii`

yamaguchi1024:
This shows up in Julien's SYRK as well. Simplify should handle value range analysis.

yamaguchi1024:
Showed up when Samir tried to use `stage_mem` over `stage_assn` and `lift_alloc`
```
for jo in seq(0, 4):
    for ji in seq(0, 16):
        C_reg[i, (jo * 16 + ji) / 16, (jo * 16 + ji) % 16] = _
```
which need to be simplified to:
```
for jo in seq(0, 4):
    for ji in seq(0, 16):
        C_reg[i, jo , ji] = _
```

PR #313, opened 2023-02-02
Div mod simplify in index expressions

SamirDroubi:


SamirDroubi:
Fixes #312

src/exo/LoopIR_scheduling.py
```diff
@@ -3139,6 +3287,20 @@ def check_quot(const, div):
 
         return check_quot(quot.lhs, quot.rhs) or check_quot(quot.rhs, quot.lhs)
 
+    @staticmethod
+    def simplify_div(e):
```
yamaguchi1024, line 3291: What is this? Do you use it?

src/exo/range_analysis.py
```diff
@@ -0,0 +1,82 @@
+from .LoopIR import LoopIR
+
+
+class AffineIndexRangeAnalysis:
```
yamaguchi1024, line 4: This class needs a comment here in explaining what it does.

src/exo/LoopIR_scheduling.py
SamirDroubi: Done

src/exo/range_analysis.py
SamirDroubi: Done

yamaguchi1024 (approved):

PR #315, opened 2023-02-02
Make cut_loop work on loops with a variable upper bound

### LINKED ISSUE #235: partition_loop() should work on loops with a variable upper bound
Title. Right now, it only works on loops that have a constant upper bound such as 
`for i in par(0, 4)`

It would assist with the implementation of TRSM and TRMM if it could be changed to work on loops like
`for i in par(0, N)`

yamaguchi1024:
Closes #235


alexreinking:
This will conflict with the changes I'm making to convert the primitive directives to use cursors.

alexreinking (approved):

PR #320, opened 2023-02-02
Filter1D example and memory aware replace

yamaguchi1024:


src/exo/platforms/x86.py
```diff
@@ -52,6 +60,17 @@ def mm256_broadcast_ss(
         out[i] = val[0]
 
 
+@instr("{dst_data} = _mm512_fmadd_ps({dst_data}, {lhs_data}, {rhs_data});")
+def mm256_fmadd_ps_broadcast(
+    dst: [f32][8] @ AVX2, lhs: [f32][8] @ AVX2, rhs: [f32][1] @ DRAM
```
alexreinking, line 65: Why can't this just be `rhs : f32 @ DRAM`?

alexreinking (approved):

src/exo/platforms/x86.py
yamaguchi1024: I think we should extend the unification to be able to unify `rhs : f32`, `rhs : f32[1]`, and `rhs : [f32][1]`, but for now, having an argument as windows is more accommodating 😅

alexreinking: I figured it was a limitation in unification... just wanted to make a note of it.

PR #322, opened 2023-02-05
x86 vector instruction for select-builtin and reduction

SamirDroubi:
Useful instructions to handle vectorization of data-based control and accumulation in BLAS 1

### DIFF
--- a/src/exo/platforms/x86.py
+++ b/src/exo/platforms/x86.py
@@ -233,11 +233,9 @@ def avx2_select_ps(
     y: [f32][8] @ AVX2,
     z: [f32][8] @ AVX2,
 ):
-    """
-    WARNING: the instructions above use a lower precision
-            float32 (C float) than the implementation of
-            the builtin which uses float64 (C double)
-    """
+    # WARNING: This instruction above use a lower precision
+    #    float32 (C float) than the implementation of
+    #    the builtin which uses float64 (C double)
     assert stride(out, 0) == 1
     assert stride(x, 0) == 1
     assert stride(v, 0) == 1
--- a/src/exo/platforms/x86.py
+++ b/src/exo/platforms/x86.py
@@ -257,7 +257,7 @@ def avx2_select_ps(
     }}
     """
 )
-def avx2_accumulate_assume_associative(x: [f32][8] @ AVX2, result: f32):
+def avx2_assoc_reduce_add_ps(x: [f32][8] @ AVX2, result: f32):
     # WARNING: This instruction assumes float addition associativity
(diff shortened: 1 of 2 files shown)


tests/test_x86.py
```diff
@@ -428,3 +428,161 @@ def sgemm_full(
...
+    result = np.array([0.0], dtype=np.float32)
+    x = np.array([1, 2, 3, 1, 2, 3, 7, 2], dtype=np.float32)
+
+    fn(None, x, result)
+
+    run_and_check(x, result)
+    run_and_check(
+        np.array(np.random.rand(8), dtype=np.float32),
+        np.array([np.random.rand()], dtype=np.float32),
+    )
```
SamirDroubi, line 588: This test may not be stable since floating point addition is calculated with different associativity. Are 4 decimals of accuracy stable enough?

yamaguchi1024:
Overall looks good to me, @alexreinking can you take a look?

tests/test_x86.py
alexreinking: Just don't make it random at runtime. Pick some specific values, even if at random once.

src/exo/platforms/x86.py
```diff
@@ -216,3 +216,49 @@ def avx2_fmadd_memu_ps(dst: [f32][8] @ DRAM, val: [f32][8] @ AVX2):
...
+    {{
+        __m256 tmp = _mm256_hadd_ps({x_data}, {x_data});
+        tmp = _mm256_hadd_ps(tmp, tmp);
+        __m256 upper_bits = _mm256_castps128_ps256(_mm256_extractf128_ps(tmp, 1));
+        tmp = _mm256_add_ps(tmp, upper_bits);
+        *{result} += _mm256_cvtss_f32(tmp);
+    }}
+    """
+)
+def avx2_accumulate_assume_associative(x: [f32][8] @ AVX2, result: f32):
```
alexreinking, line 260: For consistency with the official AVX512 intrinsics, this should be named `avx2_assoc_reduce_add_ps`

alexreinking (changes requested):
Looks pretty good, just two minor changes for test stability and naming conventions.

src/exo/platforms/x86.py
SamirDroubi: Done

tests/test_x86.py
SamirDroubi: Done

alexreinking (approved):

alexreinking:
Thanks @SamirDroubi! Will auto-merge when CI passes.

PR #323, opened 2023-02-06
Convert scheduling directives to use cursor editing APIs

alexreinking:
There's a lot to do, but I'm trying to keep the sequence of diffs/commits easily understandable.

Once this is done, the next step is to wire up the forwarding functions to the provenance system.

Then we'll have an MVP for a full cursors UX.


alexreinking:
I'm going to fix up the merge conflicts and then we can merge this early.

yamaguchi1024 (approved):

yamaguchi1024 (approved):

PR #338, opened 2023-02-20
Fix assert_if and add tests

yamaguchi1024:
assert_if didn't have any test before this and was causing errors in BLAS schedules


alexreinking (approved):

alexreinking:
@yamaguchi1024 - you need to push again

yamaguchi1024:
@alexreinking ah, thanks!

PR #342, opened 2023-02-22
add neon copy and multiply instructions

Hooninator:
Add copy instruction and alias hack multiply instruction for neon. Both instructions are used in BLAS 3 to scale matrices by alpha and beta.

### DIFF
--- a/src/exo/platforms/neon.py
+++ b/src/exo/platforms/neon.py
@@ -125,9 +125,7 @@ def neon_vmul_4xf32(
 
 
 @instr("{dst_data} = {src_data};")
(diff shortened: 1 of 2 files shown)


yamaguchi1024:
I wonder why CI is not triggered, @Hooninator can you make an empty commit to trigger the CI?

yamaguchi1024 (changes requested):
This looks like a workaround for the aliasing issue? I think this code should stay in the BLAS repo, until the issue is fixed

yamaguchi1024:
Closing as addressed in #396

PR #344, opened 2023-02-26
implemented scheduling operation for distributive property

skeqiqevian:
Needs a better name, but for now I called the operation `lift_reduce_constant`.  Implements the following rewrite pattern:
```
x = 0.0
for i in _:
    x += c * y[i]
----->
x = 0.0
for i in _:
    x += y[i]
x = c * x
```
where `c` can either be a literal constant (`LoopIR.Const`) or a buffer which is read-only within the loop (`LoopIR.Read`).


skeqiqevian:
Updated to support lifting loop constants that read read-only buffer accesses (e.g. `y[i]`)

yamaguchi1024 (changes requested):
Can you use internal_cursor's replace and insert for AST rewriting?

skeqiqevian:
> Can you use internal_cursor's replace and insert for AST rewriting?

Done!

src/exo/LoopIR_scheduling.py
```diff
@@ -1493,6 +1504,194 @@ def map_e(self, e):
...
+                                ):
+                                    return False
+                            else:
+                                raise TypeError("unknown type for self.constant")
+                        else:
+                            if isinstance(s.rhs.lhs, (LoopIR.Const, LoopIR.Read)):
+                                constant = s.rhs.lhs
+                                constant_src_stmt = s
+                            else:
+                                return False
```
alexreinking, line 1610: Reduce nesting by writing:

```
if not isinstance(s.rhs.lhs, (LoopIR.Const, LoopIR.Read)):
    return False

constant = s.rhs.lhs
constant_src_stmt = s
```

Continue to apply this strategy ("return asap upon failure") to make this code more straight-line.

src/exo/LoopIR_scheduling.py
```diff
@@ -1493,6 +1504,194 @@ def map_e(self, e):
...
+    constant = None
+    constant_src_stmt = None
+
+    for (name, typ) in get_reads_of_stmts(loop.body):
+        if assign_s.name == name and assign_s.type == typ:
+            raise SchedulingError(
+                "cannot lift constant because the buffer is read in the loop body"
+            )
+
+    def check_only_scaled_reduces(stmts):
```
alexreinking, line 1568: This is way too deeply nested for a local function. It should be rewritten to either return:

1. `None` upon failure, or
2. `constant` upon success

The `constant_src_stmt` variable is only used inside this function and doesn't need to be `nonlocal`.

It's also quite complex with nested ifs that drop-off in an effort to do nothing. This sort of thing needs comments or to be restructured.

src/exo/LoopIR_scheduling.py
```diff
@@ -1493,6 +1504,194 @@ def map_e(self, e):
...
+            if constant.name == name and constant.type == typ:
+                raise SchedulingError(
+                    "cannot lift constant because it is a buffer that is written in the loop body"
+                )
+
+    ir, fwd = assign_cursor.get_root(), lambda x: x
+
+    def lift_constant(stmts):
+        nonlocal ir
+        nonlocal fwd
```
alexreinking, line 1648: This is a good use of `nonlocal` :+1:

src/exo/LoopIR_scheduling.py
```diff
@@ -1493,6 +1504,194 @@ def map_e(self, e):
...
+                            assign_s.type,
+                            assign_s.srcinfo,
+                        ),
+                        assign_s.type,
+                        assign_s.srcinfo,
+                    )
+                )
+            ]
+        )
+    )
```
alexreinking, line 1689: Use local variables with descriptive names to avoid this sort of thing The code above is awkward to read.

src/exo/LoopIR_scheduling.py
```diff
@@ -1493,6 +1504,194 @@ def map_e(self, e):
...
+        for sc in stmts:
+            s = sc._node
+            if isinstance(s, LoopIR.Reduce):
+                if s.name == assign_s.name and same_write_dest(orig_proc, assign_s, s):
+                    assert (
+                        isinstance(s.rhs, LoopIR.BinOp)
+                        and s.rhs.op == "*"
+                        and isinstance(s.rhs.lhs, (LoopIR.Const, LoopIR.Read))
+                    )
+                    ir, fwd_repl = fwd(sc)._replace([s.update(rhs=s.rhs.rhs)])
```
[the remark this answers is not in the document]
skeqiqevian: @alexreinking I don't see a good way to get the cursor to the RHS using the current internal cursor functions. `children()` seems the most promising. If I have a statement `x += 1.0 * blah`, I could use `next(sc.children())` to get a cursor to the RHS. However, this wouldn't work if the reduce statement also had an index, e.g. `x[0] += 1.0 * blah`. I feel like I would need to use the `_children_from_attrs` function directly.

src/exo/LoopIR_scheduling.py
skeqiqevian: I restructured the code significantly, hopefully its cleaner now.

yamaguchi1024 (approved):

alexreinking (approved):

PR #354, opened 2023-03-08
rewrote fuse_if with internal cursors

skeqiqevian:


src/exo/LoopIR_scheduling.py
```diff
@@ -2625,45 +2625,26 @@ def DoFuseLoop(proc_cursor, f_cursor, s_cursor):
...
-                body1 = if1.body
-                body2 = if2.body
-                orelse1 = if1.orelse
-                orelse2 = if2.orelse
-                ifstmt = LoopIR.If(
-                    cond, body1 + body2, orelse1 + orelse2, None, if1.srcinfo
-                )
+    if1 = f_cursor._node
+    if2 = s_cursor._node
+    Check_ExprEqvInContext(proc_cursor._node, if1.cond, [if1], if2.cond, [if2])
```
alexreinking, line 2636: You can use `f_cursor._root` to get the LoopIR.proc corresponding to the root of the tree.

Then you can drop the proc_cursor argument.

alexreinking (changes requested):
Nice! Just one improvement

alexreinking:
> +20 -39

Negative PRs are the best PRs 😀

src/exo/LoopIR_scheduling.py
```diff
@@ -2596,8 +2596,8 @@ def map_s(self, sc):
         return super().map_s(sc)
 
 
-def DoFuseLoop(proc_cursor, f_cursor, s_cursor):
-    proc = proc_cursor._node
+def DoFuseLoop(f_cursor, s_cursor):
+    proc = f_cursor.get_root()
```
alexreinking, line 2600: Thanks

src/exo/LoopIR_scheduling.py
```diff
@@ -2625,45 +2625,27 @@ def DoFuseLoop(proc_cursor, f_cursor, s_cursor):
...
+    cond = if1.cond
+    body1 = if1.body
+    body2 = if2.body
+    orelse1 = if1.orelse
+    orelse2 = if2.orelse
+    ifstmt = LoopIR.If(cond, body1 + body2, orelse1 + orelse2, None, if1.srcinfo)
 
-        # if we reached this point, we didn't find the if statement
-        return super().map_stmts(stmts)
+    ir, fwd = f_cursor._delete()
```
alexreinking, line 2646: There's definitely a missing compose here.

Find a test case that breaks it and update the code here.

alexreinking (changes requested):

alexreinking (approved):

PR #356, opened 2023-03-09
Simpler AST Traversal

skeqiqevian:
Was thinking about how to rewrite some of the more complicated scheduling ops with internal cursors. Since we're trying to move away from the `Cursor_Rewrite` but still want AST traversal helpers, I wrote a fairly simple `WalkAST` class. It's already been helpful - I rewrote some existing helper functions in terms of it.

### DIFF
--- a/src/exo/LoopIR_scheduling.py
+++ b/src/exo/LoopIR_scheduling.py
@@ -176,6 +176,84 @@ class Cursor_Rewrite(LoopIR_Rewrite):
         return None
 
 
+class WalkAST:
+    """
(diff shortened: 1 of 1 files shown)


skeqiqevian:
`LoopIR_Do` already supplies the desired functionality needed for get_reads/get_writes. However, since `LoopIR_Do` operates on `LoopIR` rather than `Cursors`, it wouldn't be able to support cursor editing passes over the AST. For now, I'll try to use `pattern_match` to do what I want, but I may revisit this idea later.

PR #362, opened 2023-03-15
rewrote reuse_buffer with internal cursors

skeqiqevian:
In addition, made some changes to move cursor-specific code into `_replace_pats` functions so that I don't have to simplify the `mk_read` and `mk_write` replacement functions I have to write each time.


skeqiqevian:
Merged already

PR #367, opened 2023-03-20
rewrote write_config with internal cursors

skeqiqevian:


src/exo/API_scheduling.py
```diff
@@ -1066,14 +1066,10 @@ def write_config(proc, gap_cursor, config, field, rhs):
...
-    stmt = stmtc._impl
 
-    proc_c = ic.Cursor.create(proc)
-    rewrite_pass = scheduling.DoConfigWrite(
-        proc_c, stmt, config, field, rhs, before=before
-    )
-    mod_config = rewrite_pass.mod_eq()
-    return rewrite_pass.result(mod_config=mod_config)
+    stmt = stmtc._impl
+    (ir, cfg), _fwd = scheduling.DoConfigWrite(stmt, config, field, rhs, before=before)
```
alexreinking, line 1071: If you need to return extra information from a LoopIR directive, please do so by adding additional tuple values _after_ `ir, _fwd` and not nesting them. That way we can have robust (internal) metaprogramming later.

alexreinking (changes requested):

src/exo/API_scheduling.py
skeqiqevian: Oh ok, I think I did that because some existing scheduling operations followed that convention. I'll fix it!

alexreinking (changes requested):

alexreinking (approved):

PR #370, opened 2023-03-27
Basic cursor forwarding at API level

alexreinking:


yamaguchi1024 (approved):

skeqiqevian (approved):
Makes sense to me! Is there a reason we didn't include `delete_config`?

alexreinking:
> Makes sense to me! Is there a reason we didn't include `delete_config`?

I think I just missed it lol

PR #375, opened 2023-04-04
Fix forwarding composition in divide_dim

alexreinking:
Forwarding functions weren't being composed... I should have caught this in review.

### DIFF
--- a/src/exo/LoopIR_scheduling.py
+++ b/src/exo/LoopIR_scheduling.py
@@ -1708,11 +1708,16 @@ def DoDivideDim(alloc_cursor, dim_idx, quotient):
         except ic.InvalidCursorError:
             break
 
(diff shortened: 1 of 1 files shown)


yamaguchi1024 (approved):

skeqiqevian:
I'm not sure this is right. `_replace_pats` updates the `fwd` function and returns the updated version, so there's no need for additional composition?

alexreinking:
> I'm not sure this is right. `_replace_pats` updates the `fwd` function and returns the updated version, so there's no need for additional composition?

That's right, and was fixed in #376 -- the issue is that `_replace_pats_stmts` didn't apply the forwarding function internally. Or rather, things were type-incorrect and the function signatures were inconsistent.

PR #378, opened 2023-04-05
filter1D cursor forwarding example, and some fixes in pretty printing

yamaguchi1024:
Pretty printing is exposed to Cursor API

### DIFF
--- a/src/exo/LoopIR_pprint.py
+++ b/src/exo/LoopIR_pprint.py
@@ -677,8 +677,7 @@ def _print_cursor_stmt(
 
     else:
         lines = _print_stmt(stmt, env, indent)
-
-    if isinstance(target, Node) and isinstance(cur, Node) and cur._node == target._node:
+    if cur == target:
         lines[0] = f"{lines[0]}  # <-- NODE"
 
     return lines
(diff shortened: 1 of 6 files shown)


src/exo/LoopIR_pprint.py
```diff
@@ -677,7 +677,8 @@ def _print_cursor_stmt(
 
     else:
         lines = _print_stmt(stmt, env, indent)
-    if cur == target:
+
+    if isinstance(target, Node) and isinstance(cur, Node) and cur._node == target._node:
```
alexreinking, line 681: Why was this changed? Is there a test case that demonstrates a need for this?

alexreinking (changes requested):

src/exo/LoopIR_pprint.py
yamaguchi1024: cur and target were always not equal even though they were pointing to the same node. So "# <-- NODE" was never added to the output code.

yamaguchi1024: @alexreinking Sorry, my bad. I thought this line was failing the test I wanted to add, but it wasn't related. I think this was just some leftover from debugging. Thanks for catching this!

PR #379, opened 2023-04-05
Attempt to add BLAS correctness test to Exo repo's CI

yamaguchi1024:


.github/workflows/blas.yml
```diff
@@ -0,0 +1,64 @@
+name: CI
+on:
+  push:
+    branches: [ 'master' ]
+  pull_request:
+jobs:
+  build:
+    strategy:
```
alexreinking, line 8: The current name for this workflow and job collides with our existing CI.

```suggestion
name: BLAS
on:
  push:
    branches: [ 'master' ]
  pull_request:
jobs:
  testing:
    strategy:
```

alexreinking (changes requested):

PR #382, opened 2023-04-05
Makes cursor replaces less destructive

skeqiqevian:
Updates _replace_pats and _replace_pats_stmts to allow replacing attributes rather than the entire node. Modifies several scheduling operations accordingly.

### DIFF
--- a/src/exo/LoopIR_scheduling.py
+++ b/src/exo/LoopIR_scheduling.py
@@ -1420,7 +1420,7 @@ def DoExpandDim(alloc_cursor, alloc_dim, indexing):
     new_typ = T.Tensor(new_rngs, False, basetyp)
     new_alloc = alloc_s.update(type=new_typ)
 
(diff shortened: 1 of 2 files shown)


src/exo/LoopIR_scheduling.py
```diff
@@ -235,7 +235,28 @@ def _compose(f, g):
...
+        return ir, fwd_s
+    else:
+        if (
+            isinstance(c, ic.Block)
+            and not isinstance(c_repl, list)
+            or isinstance(c, ic.Node)
+            and c.get_index() is not None
+        ):
+            c_repl = [c_repl]
+        return c._replace(c_repl)
```
alexreinking, line 256: This seems like something that should ultimately be folded into the design of the internal cursors. But it's okay for now, since it appears to work.

alexreinking (approved):

PR #388, opened 2023-04-07
Fix forwarding bugs

skeqiqevian:


yamaguchi1024 (approved):

alexreinking (approved):

skeqiqevian:
@alexreinking I added two test cases where cursor forwarding doesn't seem to work as we would expect. It'd be great if you could take a look at it. It seems to be an issue with how `proc` equivalence is handled because the underlying LoopIRs seem to be the same.

yamaguchi1024:
@skeqiqevian Isn't that a separate issue? Can you merge this list change as is, and open the new PR separately for the failing test case?

yamaguchi1024:
@alexreinking This was the similar LoopIR.proc v.s. Procedure thing that we saw on Wednesday. This PR is ready for review; could you take a look?

PR #389, opened 2023-04-08
implemented simplify with internal cursors

skeqiqevian:
Only makes replaces at the top-level expression. Two reasons for this choice:
 - We don't have to worry about nested replaces causing the fwding function to be super long
 - Hard to modify `Cursor_Rewrite` to navigate with expr cursors unless we have introspection (pointers to proc args preds)


yamaguchi1024:
@skeqiqevian maybe it's worth using forwarding in simplify() in test_schedules.py for more test coverage?

skeqiqevian:
> @skeqiqevian maybe it's worth using forwarding in simplify() in test_schedules.py for more test coverage?

OK, I'll do that once #388 is merged.

yamaguchi1024:
@skeqiqevian merged it

yamaguchi1024 (approved):

PR #398, opened 2023-04-11
Implement autofission with cursors

yamaguchi1024:
We don't have time to debug this right now, but this is an analysis error from the new effect-check v.s. old fission checks.


yamaguchi1024:
This PR should work after fixing various analysis bugs in #578

yamaguchi1024:
Closing. It's probably worth revisiting this after merging the abstract interpretation based analysis, but this particular PR is stale.

PR #402, opened 2023-04-11
Fix new effectcheck bug, disable some checks in gemmini

yamaguchi1024:


yamaguchi1024:
FYI this PR makes the compile time very slow

PR #407, opened 2023-04-12
rewrite rearrange_dim with internal cursors

skeqiqevian:


src/exo/LoopIR_scheduling.py
```diff
@@ -1461,6 +1549,7 @@ def map_s(self, sc):
 
         # Adjust the use-site
         if isinstance(s, (LoopIR.Assign, LoopIR.Reduce)):
+            print(s.name._id, self.should_permute(s.name))
```
yamaguchi1024, line 1558:
```suggestion
```

yamaguchi1024 (changes requested):

src/exo/LoopIR_scheduling.py
```diff
@@ -1345,7 +1349,94 @@ def mk_write(c):
...
+        if not rd.name in all_permute:
+            return None
+
+        if isinstance(rd, LoopIR.WindowExpr) and not check_permute_window(
+            rd.name, rd.idx
+        ):
+            raise SchedulingError(
+                f"Permuting the window expression at {rd.srcinfo} "
+                f"would change the meaning of the window; "
+                f"propogating dimension rearrangement through "
```
alexreinking, line 1399:
```suggestion
                f"propagating dimension rearrangement through "
```
Typo.

alexreinking (changes requested):
Looks good, but need to correct a typo in an error message.

PR #417, opened 2023-04-26
Reduce cursor invalidation

skeqiqevian:
- ProductLoop - make outer loop cursor forward to the the new loop
- LiftScope - avoid invalidating cursors to the `if` and `for` stmts when lifting a for loop out of an if statement
- RemoveLoop - avoid invalidating cursors to the body
- FuseLoop - avoid invalidaing cursors to the 2nd loop's body

### DIFF
--- a/src/exo/LoopIR.py
+++ b/src/exo/LoopIR.py
@@ -580,36 +580,18 @@ class LoopIR_Rewrite:
     def apply_t(self, old):
         return self.map_t(old) or old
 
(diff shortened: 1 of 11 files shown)


yamaguchi1024 (approved):

yamaguchi1024 (changes requested):
@skeqiqevian Wait, actually it seems there is compilation failure from BLAS

yamaguchi1024 (approved):

PR #423, opened 2023-05-01
CIR for index access simplification

yamaguchi1024:


src/exo/memory.py
```diff
@@ -58,6 +58,34 @@ class MemGenError(Exception):
...
+            return i
+        if i == "1":
+            return s
+
+        if len(s) == 1:
+            return f"({i}) * {s}"
+        else:
+            return f"({i}) * ({s})"
+
+    offset = index_expr(indices[0], strides[0])
```
skeqiqevian, line 76: I think we can rewrite this to map all the indices, then filter out the "", and then either return `"+".join(exprs)` if `len(exprs) > 1` and `"0"` otherwise.

skeqiqevian (approved):
Looks good, just had a small nitpick. 

Also for future note: it's a little weird that the simplification behavior is different for `window` (since we have to do that simplification through memories.py), but I can't think of a good solution to that right now.

src/exo/memory.py
yamaguchi1024: True, will do that!

PR #425, opened 2023-05-03
Allow loops to have arbitrary expressions as `lo`

skeqiqevian:


yamaguchi1024 (changes requested):
@skeqiqevian 
Sorry for being so slow reviewing this! I just wanted to play around and write more tests. The only bug I found is in unroll_loop, when you unroll
```
for i in seq(3, 10):
    tmp[i] = A[i]
```
it generates
```
for i in seq(3, 10):
    tmp[0] = A[0]
    ....
```
instead of 
```
for i in seq(3, 10):
    tmp[3] = A[3]
    ....
```
The failing test case is in `test_schedules::test_simple_unroll2`

yamaguchi1024 (approved):
LGTM, thanks!!

PR #427, opened 2023-05-04
Add alias check. Fixes #275.

### LINKED ISSUE #275: We should check Check_Aliasing from the frontend typecheck

rachitnigam:


yamaguchi1024 (approved):

yamaguchi1024:
@rachitnigam It looks like some tests are failing because their algorithm has aliasing. Could you modify the tests and golden output so that they don't have aliasing?

rachitnigam:
Hey @yamaguchi1024! I’ll get this done early next week. I got pulled into a deadline so a little busy right now

yamaguchi1024:
@rachitnigam No worries & no rush, good luck for the deadline!

yamaguchi1024:
Actually, it seems what we want is to allow aliasing in general, but check if the argument can be aliased in the backend and add `__restrict__` if not.

gilbo:
If you want to allow aliasing you need to account for it everywhere in the program analysis…On May 9, 2023, at 8:05 AM, Yuka Ikarashi ***@***.***> wrote:﻿
Actually, it seems what we want is to allow aliasing in general, but check if the argument can be aliased in the backend and add __restrict__ if not.

PR #428, opened 2023-05-04
ArgCursor::mem assert bug fix

SamirDroubi:


yamaguchi1024:
@SamirDroubi how about changing it to is_tensor_or_window() instead of killing it altogether?

SamirDroubi:
@yamaguchi1024  What about scalars? You should be able to see what the memory type is for type, similar to `AllocCursor` https://github.com/exo-lang/exo/blob/72e76d2d3b730a1d526886e04bc3c3a848fdde37/src/exo/API_cursors.py#L588

yamaguchi1024:
@SamirDroubi True, what we want instead is `assert not self.type.is_indexable()` ?

yamaguchi1024 (approved):

PR #434, opened 2023-05-05
Fix typo in new_eff

yamaguchi1024:


skeqiqevian (approved):
This change is fine. We should probably add a test that hits this path though...

PR #439, opened 2023-05-31
Propagate InferEffects to subprocedure calls

yamaguchi1024:
Fixes a bug where the changes in `foo`'s effects were not propagated after partial eval, resulting in the effectcheck error in `bar`.

### DIFF
--- a/src/exo/API.py
+++ b/src/exo/API.py
@@ -238,9 +238,11 @@ class Procedure(ProcedureBase):
         return self._loopir_proc.name
 
     def show_effects(self):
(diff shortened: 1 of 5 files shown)


skeqiqevian (changes requested):
I’m not sure this solves the root problem. For instance, if you were to add `print(foo.show_effects())` after the `partial_eval` call, it would give
```
Writes:
    { A : (i,j) for (i,j) in Z if
    0 <= i and i < n and (0 <= j and j < m) }
```
which is wrong. I think the root issue is that `partial_eval` doesn’t update effects?

skeqiqevian (approved):

PR #442, opened 2023-06-02
Fix a typo in new_analysis core

yamaguchi1024:


skeqiqevian (approved):
LGTM. We should also fix the typo in ` _get_real_const` (line 830), which uses `name` instead of `val`...

yamaguchi1024:
@skeqiqevian Amazing catch! Will fix it. Why was this working tho..

PR #468, opened 2023-07-10
Rework the Scheduling Example

rachitnigam:
I was reading the scheduling example and found it a bit magical so trying to rewrite it a bit and add more explanations for how exactly the scheduling operators work and what the goals are. Not ready for review yet so I'll ping when ready.


gilbo:
First off, thanks so much for rewriting this.  I have a tendency to just plow through with a ton of nitpicky criticisms, so please take it as anything other than constructive.  This is so much better than what we had before.

I'm reading through the rendered version alone for comments here...
* The inline LaTeX appears to be broken?
* It may be worth motivating _why_ we would want to write a 6x16xK dimension matrix multiplication "microkernel".  You can refer to the BLIS project here and say that one popular approach to implementing high performance BLAS libraries is built around specializing these kinds of inner-loop routines for different architectures.  Creating and tuning these sorts of high-performance micro-kernels up to near peak machine throughput is the most basic step in creating high performance GEMMs and similar routines.
* You can probably assert that most vectorizing compilers will not correctly optimize this code.  JRK may have useful opinions on how braggadocio you can safely get here, but it doesn't hurt to turn it up a bit within reason.
* "Also, just to keep things easy to follow, we're going to rename our kernel" - a slight tweak, but we recommend that users always rename any variant of a procedure that they are trying to schedule.  This prevents name conflicts if and when you try to output multiple versions in the same C-library later on.
* We have called things like `reorder_loops` by the term "scheduling operators."  I'm not sure if you want to use that term or not. ¯\_(ツ)_/¯  "scheduling command" sounds fine too.
* You may want to mention that "proc" is short for "Procedure"
* Your expansion of the `'j k'` nested loop pattern is slightly wrong.  The first line should not have a `_` after the loop.  The body is the second loop!
* "Finally the `print(avx)` shows us..."  You may want to suggest that the best way to do scheduling is ABP: Always Be Printing.
* "...represented by 12 8-wide vectors..." -> "stored in"?
* "Even though we're streaming..."  You might want to consider referring to this as "register blocking" here.  Readers may have less trouble Googling that term and will see that it's a standard optimization strategy for the inner-most loops of programs.
* In "Vectorizing the Output" and continuing from there, you leave off printing the output.  I realize it's a very verbose thing to do, but it might be a good idea none-the-less, given that the example writeup is already long?
* "`repeat` which applies a scheduling operator till no new matches are found"  A slight correction is to say that it keeps applying until the operation _fails_ whether because of a pattern match failing or for whatever other reason.

Btw, for what K and machine specs are you reporting those perf numbers?  It might be a good idea to make sure we're not doing something wrong performance-wise.

jrk:
Dumb thing this made me notice (but was there all along): is it really only a 2.2x speedup over the trivial triple for loop?

rachitnigam:
@gilbo Thanks for the comments! I'll address them soon. @jrk the perf numbers are from the old text so I haven't validated them. After @gilbo's comment, I was just planning to remove the output since I haven't been able to validate it on my M1 machine

rachitnigam:
Okay, addressed the comments! As far as the number go, it seems that [Rosetta 2 cannot emulate AVX2 instructions](https://www.computerworld.com/article/3597949/everything-you-need-to-know-about-rosetta-2-on-apple-silicon-macs.html) so I can get rid of the output from running that command for now and open and issue to add real numbers from running on an x86 machine in a future PR?

rachitnigam:
@gilbo if the changes look good, we should merge

examples/README.md
```diff
@@ -1,214 +1,267 @@
...
             for k in seq(0, K):
                 C[i, j] += A[i, k] * B[k, j]
+
+print(rank_k_reduce_6x16)
 ```
 
-Next, please uncomment the code in the first block by deleting the multi-line string
-markers (`"""`). Now, you will see that `stage_mem()` stages `C` to a buffer
-called `C_reg`.
+This implements matrix multiplication between a $6\times K$ and a $K \times 16$. This might initially seem like a contrived example: the input sizes of the matrices are non-standard and only one of the dimensions is kept as an argument. However, it is precisely such *microkernels* that lie at the heart of the most computationally intensive tasks. For example, the [BLAS][] defines several useful linear algebra kernels which are often specialized for target architectures and even matrix sizes for specific workloads. The goal of Exo is to make this specialization process dramatically easier.
```
gilbo, line 29: Alternate text starting at "However":
> However, these *microkernels* function as the inner loops of highly optimized linear algebra computations.  For example, [BLIS](https://github.com/flame/blis) (an open-source BLAS library) is architected around re-implementing such microkernels for each new target architecture that they support.  The goal of Exo is to make this specialization process dramatically easier.

examples/README.md
```diff
@@ -1,214 +1,267 @@
...
-def rank_k_reduce_6x16_scheduled(K: size, C: f32[6, 16] @ DRAM,
-                                 A: f32[6, K] @ DRAM, B: f32[K, 16] @ DRAM):
-    for i in seq(0, 6):
-        for j in seq(0, 16):
-            for k in seq(0, K):
-                C_reg: f32 @ AVX2
-                C_reg = C[i, j]
-                C_reg += A[i, k] * B[k, j]
-                C[i, j] = C_reg
+For our example, we want to specialize the kernel to use the AVX2 instructions and, it is likely the case that a vectorizing compiler cannot automatically transform this kernel.
```
gilbo, line 31: Can just break into two sentences at "and"

examples/README.md
```diff
@@ -1,214 +1,267 @@
...
 $ exocc -o . --stem avx2_matmul x86_matmul.py
 $ gcc -o avx2_matmul -march=native main.c avx2_matmul.c
 ```
 
 It should generate something like:
 
-```
+```sh
 $ ./avx2_matmul
 Time taken for original matmul: 0 seconds 649 milliseconds
```
gilbo, line 261: I think we decided to cut the times here?

gilbo (changes requested):
Some very slight tweaks.  Main thing is we want to pull out those performance numbers before merging the PR

rachitnigam:
@gilbo addressed comments!

gilbo (approved):

PR #470, opened 2023-07-12
Handle cursor for reorder_loops

rachitnigam:


rachitnigam:
@gilbo @yamaguchi1024 what would be a good place to put a test for this? Essentially, we want to make sure that the following works:
```py
reorder_loops(proc, 'for i in _: _')
```
as a way to specify the cursor needed to reorder the loops

yamaguchi1024:
Closing, `bar = reorder_loops(bar, bar.find("for i in _:_"))` works

PR #491, opened 2023-09-22
Revised cut_loop and implemented shift_loop

SamirDroubi:


src/exo/LoopIR_scheduling.py
```diff
@@ -301,42 +313,74 @@ def DoReorderStmt(f_cursor, s_cursor):
...
+    assert isinstance(s, LoopIR.Seq)
+
+    try:
+        Check_IsNonNegativeExpr(
+            loop_c.get_root(),
+            [s],
+            new_lo,
+        )
+    except SchedulingError:
+        raise SchedulingError(f"Expected 0 <= `new_lo`")
```
skeqiqevian, line 359: Is there a reason we don't want negative loops?

skeqiqevian (approved):
Mostly looks good! I deleted a duplicated test and used `shift_loop` to revert the SYRK output.

src/exo/LoopIR_scheduling.py
SamirDroubi: My understanding is that so far we have been maintaining the invariant that
0 <= lo <= hi

I am not sure if changing it would break anything somewhere. It might be worth it to revise LoopIR to have actual classes with initializer to document those invariants in code as asserts.

skeqiqevian: Hmm, that might be an artifact of having lo = 0 in our previous representation of for loops. In any case, it's probably not essential to support negative values right now though.

PR #492, opened 2023-09-23
assert_if revision

SamirDroubi:


src/exo/API_scheduling.py
```diff
@@ -1942,30 +1942,25 @@ def lift_scope(proc, scope_cursor):
 
 
 @sched_op([IfCursorA, BoolA])
-def assert_if(proc, if_cursor, cond):
+def remove_if(proc, if_cursor, cond):
     """
```
yamaguchi1024, line 1946: Why does it take `cond` when it's not passed to `DoRemoveIf` ?

SamirDroubi: Ahhh I forgot to deprecate the API parameter.

SamirDroubi: Done

yamaguchi1024: LGTM, I'll delegate it to @skeqiqevian to approve

skeqiqevian (approved):
Looks good to me!

PR #522, opened 2023-10-23
More general bounds inference

skeqiqevian:


apps/blur/blur.py
```diff
@@ -45,10 +45,10 @@ def prod_inline(p):
 
     c_bounds = (0, "i", 0, 1)
     p_bounds = (0, "i", 0, 5)
```
SamirDroubi, line 47: Dead code

apps/blur/blur.py
```diff
@@ -45,10 +45,10 @@ def prod_inline(p):
...
-    p = fuse_at(p, "f", "g", p.find_loop("i #1"), c_bounds, p_bounds)
+    p = fuse_at(p, "f", "g", p.find_loop("i #1"))
 
     loop = p.find_loop("i")
-    p = store_at(p, "f", "g", loop, p_bounds)
+    p = store_at(p, "f", "g", loop)
 
     c_bounds_2 = (1, "j", 0, 1)
     p_bounds_2 = (1, "j", 0, 1)
```
SamirDroubi, line 54: Dead code

src/exo/stdlib/scheduling.py
```diff
@@ -348,56 +351,48 @@ def lift_if(proc, cursor, n_lifts=1):
...
-
-def fuse_at(
-    proc, producer, consumer, loop, consumer_bounds, producer_bounds, hardcode=1
-):
+def fuse_at(proc, producer, consumer, loop):
     """
     This version of compute_at will only go down one-level of for loops
 
     TODO: bounds
      - currently assumes that bounds is of the form [0, 1, ..., n-1]
```
SamirDroubi, line 359: It seems like this TODO can also be removed since fuse_at no longer takes any bounds arguments.

src/exo/stdlib/scheduling.py
```diff
@@ -94,10 +94,13 @@
     autolift_alloc,
 )
 
+
 from .analysis import (
     check_call_mem_types,
 )
 
+from exo.range_analysis import IndexRange, bounds_inference, get_affected_idxs
+
```
SamirDroubi, line 103: We need to be careful about illegally importing compiler code in the standard library.

Perhpas these functions should have an API wrapper similar to how we have an API wrapper for scheduling ops in API_scheduling.py. We should also make sure to maintain a similar calling/naming convention. I am not sure what's a good name for it? API_introsepction.py, API_analysis.py?

src/exo/range_analysis.py
```diff
@@ -1,104 +1,171 @@
...
-
-    Any range, in the environment mapping or the result of the
-    analysis is either:
-        A tuple `T` of length 2 s.t. it represents the range
-        `[T[0], T[1]]` (both inclusive).
-        If `T[0]` or `T[1]` is `None` it represents no knowledge
-        of the value of that side of the range.
+    Represents a range of possible values between [base + lo, base + hi]
+        - [base] contains all expressions that we don't want to bounds infer (e.g. free
+        variables). If base is None, it signifies a base of 0.
```
SamirDroubi, line 19: I think representing the base = None as LoopIR.Const(0) might simplify a lot of the code later.

src/exo/range_analysis.py
```diff
@@ -1,104 +1,171 @@
...
+        # +1 because bounds are inclusive
+        return self.hi - self.lo + 1
+
+    def __str__(self):
+        base = "0" if self.base is None else str(self.base)
+        lo = "-inf" if self.lo is None else str(self.lo)
+        hi = "inf" if self.hi is None else str(self.hi)
+        return f"({base}, {lo}, {hi})"
+
+    def __add__(self, other: int | IndexRange) -> IndexRange:
```
SamirDroubi, line 53: Can we add assert to check the types of all these operators since some are expecting IndexRange, some are expecting integers, and some accept either? Or would these types hint add any sort of runtime checks, I am not familiar with that feature.

SamirDroubi (changes requested):
I reviewed a good part of this PR and left some comments. I still need to go through the reaminder of the functions in range_anaylsis.py and the testing code.

src/exo/range_analysis.py
```diff
@@ -1,104 +1,171 @@
...
-    analysis is either:
-        A tuple `T` of length 2 s.t. it represents the range
-        `[T[0], T[1]]` (both inclusive).
-        If `T[0]` or `T[1]` is `None` it represents no knowledge
-        of the value of that side of the range.
+    Represents a range of possible values between [base + lo, base + hi]
+        - [base] contains all expressions that we don't want to bounds infer (e.g. free
+        variables). If base is None, it signifies a base of 0.
+        - [lo] and [hi] are bounds. If either is None, it means that there is no
+        constant bound
```
SamirDroubi, line 21: I made the mistake of starting the convention of using None as an unkown. I think it might be a good idea to implement a type a that is basically Integers+Unknwon and hide all the messiness behind the implementation of that type's operator. I can work on it tomorrow.

src/exo/stdlib/scheduling.py
skeqiqevian: No, it still makes the implicit assumption that the region is contiguous because it generates for loops from `lo` to `hi` without regard for whether all the values from `[lo, hi]` are accessed.

src/exo/range_analysis.py
skeqiqevian: Isn't that essentially what I have with `Optional[int]`?

src/exo/range_analysis.py
skeqiqevian: The problem with this approach is that we would either (a) generate a lot of 0s from all the interval merging and yield expressions like `0 * 0 + 0 + i` or (b) effectively have to check against `LoopIR.Const(0, blah)` to avoid the issue in (a), which is essentially what I'm already doing with `None` now anyways.

src/exo/range_analysis.py
SamirDroubi: Yes, but I am thinking we can then do the ranges merging without having to do all the is None checks

src/exo/stdlib/scheduling.py
skeqiqevian: I will fix this in a future PR to better separate user-level and compiler-level bounds inference.

src/exo/range_analysis.py
skeqiqevian: I think we can probably just change that later? I think that change is fairly orthogonal to everything we need for the paper deadline, and may possibly be irrelevant if we ultimately replace some of this with abstract interpretation.

SamirDroubi (approved):

PR #526, opened 2023-11-02
Adds parallel loops and scheduling op to parallelize loops

skeqiqevian:
This is motivated by the effort to support Halide's `compute_at` and `store_at` commands in Exo. The observation is that for blur, these operations are only useful when used **with** parallel threads. 

As part of this chance, we renamed `Seq` to `For` so that we can differentiate between sequential and parallel for loops. This will also require modifying the BLAS repository accordingly to reflect that. 

Some future considerations: 
- Do we ever want a scheduling operation to revert back to sequential loops? 
- Different types of parallelism that we want to represent – the current one just generates `#pragma omp parallel for`.


SamirDroubi:
do you have any thoughts about things that needs to be handled at the boundary between Exo and C once we depend on OpenMP? I am thinking omp isn't part of the standard, so you might want to check if the pragma is available/not available in the target C compiler from Exo or perhaps you can just rely on a failure in the downstream compiler. Although what if you ask Exo to simply emit C code, then you wouldn't be able to check for availability. One option is to have some C macros that would check if the pragma is available or not, and if not potentially generate C loops without the pragma. Although you might actually want downstream C compilation to just fail rather than silently default to regular loops.

src/exo/LoopIR_scheduling.py
```diff
@@ -320,6 +321,12 @@ def DoReorderStmt(f_cursor, s_cursor):
     return ir, fwd
 
 
+def DoParallelizeLoop(loop_cursor):
+    # TODO: Check commutativiity of loop body
+    Check_ParallelizeLoop(loop_cursor.get_root(), loop_cursor._node)
+    return loop_cursor._child_node("config")._replace(LoopIR.Par())
```
yamaguchi1024, line 327: I stared at this for 30 seconds wondering why configuration is replaced here, can we use other word for "loop_config"? How about "is_par" or "loop_mode"?

yamaguchi1024: config will be sooo confusing in Exo

SamirDroubi: Can't we rely on inheritance here?
For = Seq()
          | Par()
          attributes(sym iter, expr lo, expr hi, stmt* body, loop_config config)

This is also to simplify code. I would imagine being able to immediately write:
`isinstance(node, Par)` when I want to check if some random node is a parallel loop as opposed to writing `isinstance(node, For)` and `isinstance(node.config, Par)`

skeqiqevian: I don't believe ASDL allows for that kind of inheritance, based on what I could tell from its grammar.

src/exo/LoopIR_scheduling.py
```diff
@@ -320,6 +321,12 @@ def DoReorderStmt(f_cursor, s_cursor):
     return ir, fwd
 
 
+def DoParallelizeLoop(loop_cursor):
```
yamaguchi1024, line 324: I still think parallel check shouldn't be a part of the rewriting process but should be a part of backend checks (same as memory and precision), because even if you check the safety in parallelize_loop , other scheduling primitives do not know that the loop is parallel and can rewrite the par loop to introduce the loop dependencies and break the safety.

gilbo:
100% that the safety of parallel loops is a post-scheduling check.  That was the design that was green lighted.  You will have a lot of trouble if you try to check safety of parallelization in any other way.

skeqiqevian:
Agreed about making it a post-scheduling pass @yamaguchi1024 and @gilbo.

skeqiqevian:
> do you have any thoughts about things that needs to be handled at the boundary between Exo and C once we depend on OpenMP? I am thinking omp isn't part of the standard, so you might want to check if the pragma is available/not available in the target C compiler from Exo or perhaps you can just rely on a failure in the downstream compiler. Although what if you ask Exo to simply emit C code, then you wouldn't be able to check for availability. One option is to have some C macros that would check if the pragma is available or not, and if not potentially generate C loops without the pragma. Although you might actually want downstream C compilation to just fail rather than silently default to regular loops.
> 
> I am not even really sure that you can check at compile-time whether a pragma exists.

Yeah, in order to compile files which use parallel for loops, we need to pass the `fopenmp` flag. I'm not sure if Exo is able to do anything about that unless you want Exo to also encompass the build system?

jrk:
FWIW, I **think** code with `#pragma omp parallel` will just silently compile (with sequential loops) in a compiler that doesn't support OpenMP (or doesn't have it turned on). I think this is fine.

src/exo/LoopIR_scheduling.py
```diff
@@ -320,6 +320,10 @@ def DoReorderStmt(f_cursor, s_cursor):
...
```
yamaguchi1024, line 323: I think it's fine for now, but we probably need a scheduling operation to turn a parallel loop into a sequential loop

yamaguchi1024 (approved):
LGTM!

yamaguchi1024:
Don't forget to change the Cursor introspection in BLAS

PR #527, opened 2023-11-06
Add guards to load/store stage in stage_mem

SamirDroubi:
* stage_mem didn't check the input window bounds.
* We add guards if neccessary to make sure load/store are always in-bounds
* This is useful for handling tail cases where you often want to round up your accesses


SamirDroubi:
Making this a draft until I make sure it compltely unblocks the other scheduling work I am doing, but it should be a useful PR regardless.

yamaguchi1024:
LGTM, but I defer this review to @skeqiqevian

skeqiqevian:
@SamirDroubi LGTM, this is a strict improvement over what we previously had. However, I just want to note that this doesn't fully fix `stage_mem`. Consider the following example: 
image
This code will clobber the original values stored at `x[3:4]` with zero/garbage values. I think to fully fix, we need better index reasoning in effects. @yamaguchi1024 this might be useful to be aware of when designing the new effects.

PR #528, opened 2023-11-08
Dram stack

skeqiqevian:
Need to support `DRAM_STACK` in addition to `DRAM_STATIC` because of parallel semantics. If we want to allocate temporary memory buffers in parallel code, `static` memory will lead to races because it isn't thread local.

Depends on #526


src/exo/LoopIR_scheduling.py
```diff
@@ -18,6 +18,7 @@
 from .new_eff import (
     SchedulingError,
     Check_ReorderStmts,
+    Check_ParallelizeLoop,
```
yamaguchi1024, line 21: This is unnecessary, right?

src/exo/LoopIR_compiler.py
```diff
@@ -397,10 +397,10 @@ def from_lines(x):
         else:
             is_public_decl = id(p) in orig_procs
 
+            p = ParallelAnalysis().run(p)
```
yamaguchi1024, line 400: Why?

skeqiqevian: As in why is it first? Or why at all?

yamaguchi1024: Why did you move it to the first?

skeqiqevian: I don't think the current effects analysis handle frees correctly, so I moved it first. We'll probably have to revisit this ordering later for GPUs, but in the current state, I don't think the order particularly matters?

yamaguchi1024: Okay yeah that's fine

src/exo/LoopIR_scheduling.py
yamaguchi1024: Do you use Check_ParallelizeLoop in LoopIR_scheduling.py? If so we shouldn't right?

yamaguchi1024 (approved):

PR #529, opened 2023-11-10
Blur sprint

skeqiqevian:
Depends on Samir's `vectorize`, will wait until that is finished before finishing this PR.


skeqiqevian:
This is finally ready for review. Note that the large sections of inserts/deletes are likely just code reorganization. There are still some TODOs written in the code in terms of the code not being general enough, but I will deal with those later when I try to write more complicated Halide schedules.

yamaguchi1024 (approved):
LGTM!

PR #530, opened 2023-11-10
Allow dividing loops by 1

SamirDroubi:
* Often when you are scheduling step by step, you shouldn't really need to divide a loop by 1 which is why this was disabled.
* As you add more automation and implement more abstraction, you want to rely on symmetries in the structure of the AST and having this disabled is diruptive.


SamirDroubi:
I test before/after simplify to make sure the division was correct before I simplify and that after simplify things I would expect to be simplified are in fact simplified

skeqiqevian (approved):
LGTM if it passes BLAS tests.

PR #531, opened 2023-11-11
Forwarding for replacing nodes

skeqiqevian:
If we're going to claim that `_replace` is different from `_insert` and `_delete`, that should be true lol. 

TODO: I will refactor the large copy and pasted code from `_local_forward` into a separate function, since it now appears three times in the code (once in the original `_local_forward`, once for `_move`, and now once for `_replace`).


skeqiqevian:
See #546 for new version.

PR #536, opened 2023-12-06
Added support for uint16 and improved `replace_all` to handle bodies with length > 1

skeqiqevian:
This commit adds a precision for `uint16_t` to Exo, which was necessary to compare against the Halide blur benchmark. I also added the corresponding AVX2 instructions that were necessary for the blur effort. One important system change is that `replace_all` is now capable of handling instructions whose proc semantics are longer than a single statement.


src/exo/stdlib/scheduling.py
```diff
@@ -269,19 +264,26 @@ def _replace_helper(proc, subprocs, mem_aware, once):
         i = 0
         while True:
             try:
+                block = proc.find(f"{pattern} #{i}").expand(0, len(body) - 1)
+                assert len(block) == len(
+                    body
+                ), "block is too close to the end of the body"
```
yamaguchi1024, line 270: @skeqiqevian 
Is this supposed to be an error when the length of a block from the pattern is less than the length of the subproc body? If so, shouldn't this be SchedulingError? Try/catching the `AssertionError` in the code below may hide various internal compiler errors

skeqiqevian: As it currently works, `expand` will automatically clamp to the end of the scope. So if `proc.find` returns a statement at the end of a body, and we try to expand it by 2, we will receive a block of size 1. Maybe this should be a type of UnificationError?

yamaguchi1024: Yeah I think `UnificationError` is reasonable here. If you change that, can you remove catching `AssertionError` from below?

yamaguchi1024 (approved):

src/exo/stdlib/scheduling.py
```diff
@@ -269,19 +264,26 @@ def _replace_helper(proc, subprocs, mem_aware, once):
...
             except (TypeError, SchedulingError) as e:
                 if "failed to find matches" in str(e):
                     break
                 raise
-            except (_UnificationError, MemoryError, NotImplementedError):
+            except (
+                _UnificationError,
+                MemoryError,
+                NotImplementedError,
+                AssertionError,
```
yamaguchi1024, line 285: Oh wait you wanna remove this

yamaguchi1024 (changes requested):
Don't catch AssertionError

yamaguchi1024 (approved):

PR #537, opened 2023-12-06
Changing shrink_dim to resize_dim

skeqiqevian:
This will make the command more useful, I think @SamirDroubi had a use case? In some sense, this encapsulates the old `bound_alloc`.


src/exo/stdlib/scheduling.py
```diff
@@ -269,19 +264,26 @@ def _replace_helper(proc, subprocs, mem_aware, once):
         i = 0
         while True:
             try:
+                block = proc.find(f"{pattern} #{i}").expand(0, len(body) - 1)
```
yamaguchi1024, line 267: Duplicate with #536

tests/test_schedules.py
```diff
@@ -401,37 +401,37 @@ def foo():
...
-    foo = shrink_dim(foo, "x", 0, 2, "n-1")
+    foo = resize_dim(foo, "x", 0, "n-3", 2)
     assert str(simplify(foo)) == golden
 
     with pytest.raises(SchedulingError, match="The buffer x is accessed out-of-bounds"):
-        foo = shrink_dim(foo, "x", 0, 2, "n-2")
+        foo = resize_dim(foo, "x", 0, "n-4", 2)
     with pytest.raises(SchedulingError, match="The buffer x is accessed out-of-bounds"):
-        foo = shrink_dim(foo, "x", 0, 3, "n-1")
+        foo = resize_dim(foo, "x", 0, "n-4", 3)
```
yamaguchi1024, line 434: Can you add a test case where `offset` is not an integer constant? (e.g., `i-5` or sth?)

skeqiqevian: I think that implicitly exists in Halide test cases, but yes I can add one here too.

src/exo/stdlib/scheduling.py
skeqiqevian: Oops.

PR #542, opened 2023-12-19
Bind expr fixes

### LINKED ISSUE #509: bind_expr Bugs
1. bind_expr doesn't accept an argument that is a curosr to an expressions, but rather it must be given a list always if we are using a cursor. This is a bug in the arguments processor. 
Bug example:
```
def test_bind_expr_issue():
    @proc
    def bar(n: size, x: f32[n], y: f32[n], z: f32[n]):
        for i in seq(0, n):
            x[i] = y[i]
            x[i] = z[i]
    
    yi_cursor = bar.find("y[_]")
    zi_cursor = bar.find("z[_]")
    bar = bind_expr(bar, [yi_cursor], "y_read")     <----  Okay
    bar = bind_expr(bar, zi_cursor, "z_read")       <----- Error
````

2. Passing a list of more than one cursor to bind_expr binds only the first cursor unless `cse` is set to True. I find this very confusing because why else would I want to pass a list of many cursors:
```
def test_bind_expr_issue():
    @proc
    def bar(n: size, x: f32[n], y: f32[n], z: f32[n]):
        for i in seq(0, n):
            x[i] = y[i]
            z[i] = y[i]
    
    yi_cursors = bar.find("y[_]", many=True)
    print(bind_expr(bar, yi_cursors, "y_read"))
    print(bind_expr(bar, yi_cursors, "y_read", cse=True))
```
```
def bar(n: size, x: f32[n] @ DRAM, y: f32[n] @ DRAM, z: f32[n] @ DRAM):
    for i in seq(0, n):
        y_read: R
        y_read = y[i]
        x[i] = y_read
        z[i] = y[i]
def bar(n: size, x: f32[n] @ DRAM, y: f32[n] @ DRAM, z: f32[n] @ DRAM):
    for i in seq(0, n):
        y_read: R
        y_read = y[i]
        x[i] = y_read
        z[i] = y_read
```
Perhpas we should internally set cse=True whenever the user passes a list of cursors. I think generally it is unclear to me what cse implies in the context of cursors and in the contex of template arguments. We should at least document this behvaior if it is what is expected.
3. bind_expr currently assigns the type R to the new allocation. I think this is a bit confusing since when I bind something I would expect the default to be the type of the original expressions, otherwise you can easily get into a situation where you have cross-precision computation even though your original computation didn't have that. Changing the type of some computation shouldn't be something that is silently done by a scheduling operations, but only by explicit direction from the processor by using set_precision. Another reason why this is confusing is that stage_mem would use the precision of the staged buffer which makes it a bit inconsinstent for bind_expr to do something else.

### PR DESCRIPTION
**Pull request by SamirDroubi**
Partially addresss #540 and fixes #509 .

### DIFF
--- a/apps/aarch64/filter1D/filter1D_cursor_forwarding.py
+++ b/apps/aarch64/filter1D/filter1D_cursor_forwarding.py
@@ -42,7 +42,6 @@ filter1D = fission(filter1D, filter1D.find("xX4[_] = _").after())
 # set memories & precision
 filter1D = set_memory(filter1D, "sum #1", Neon)
 filter1D = set_memory(filter1D, "xX4", Neon)
-filter1D = set_precision(filter1D, "xX4", "f32")
 
 # replace
 filter1D = replace(filter1D, "for outXi in _:_", neon_zero_4xf32)
--- a/tests/test_apps.py
+++ b/tests/test_apps.py
@@ -44,7 +44,7 @@ def test_gemmini_matmul(golden):
     assert _test_app(module_file) == golden
 
 
-@pytest.mark.slow
+@pytest.mark.skip(reason="")
 def test_gemmini_conv(golden):
     module_file = REPO_ROOT / "apps" / "gemmini" / "src" / "exo" / "conv.py"
     assert _test_app(module_file) == golden
--- a/tests/test_apps.py
+++ b/tests/test_apps.py
@@ -44,7 +44,7 @@ def test_gemmini_matmul(golden):
     assert _test_app(module_file) == golden
 
 
-@pytest.mark.skip(reason="")
+@pytest.mark.slow
 def test_gemmini_conv(golden):
     module_file = REPO_ROOT / "apps" / "gemmini" / "src" / "exo" / "conv.py"
     assert _test_app(module_file) == golden
--- a/apps/aarch64/filter1D/filter1D_old.py
+++ b/apps/aarch64/filter1D/filter1D_old.py
@@ -34,7 +34,6 @@ filter1D = autofission(filter1D, filter1D.find("xX4[_] = _").after())
 # set memories & precision
 filter1D = set_memory(filter1D, "sum #1", Neon)
 filter1D = set_memory(filter1D, "xX4", Neon)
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

**Comment by SamirDroubi** [comment]
It is expected that BLAS tests will fail.

File: apps/aarch64/filter1D/filter1D_cursor_forwarding.py [hunk 1]
```diff
@@ -42,7 +42,6 @@ def filter1D(ow: size, kw: size, x: f32[ow + kw - 1], y: f32[ow], w: f32[kw]):
 # set memories & precision
 filter1D = set_memory(filter1D, "sum #1", Neon)
 filter1D = set_memory(filter1D, "xX4", Neon)
-filter1D = set_precision(filter1D, "xX4", "f32")
```
yamaguchi1024, line 45: Why was this removed?

tests/test_apps.py
```diff
@@ -44,7 +44,7 @@ def test_gemmini_matmul(golden):
     assert _test_app(module_file) == golden
 
 
-@pytest.mark.slow
+@pytest.mark.skip(reason="")
```
yamaguchi1024, line 47: Gemmini conv shouldn't be skipped

SamirDroubi: My bad, I usually disable this test because it takes forever to run locally. Accidently, pushed that code.

apps/aarch64/filter1D/filter1D_cursor_forwarding.py
SamirDroubi: Since bind_expr automatically infers the type it is no longer necessary to set the precision.

tests/test_apps.py
yamaguchi1024: I think pytest took --fast option or sth like that to skip slow tests from the commandline

apps/aarch64/filter1D/filter1D_cursor_forwarding.py
yamaguchi1024: I see, nice!

yamaguchi1024 (approved):

PR #543, opened 2023-12-19
Pattern matching on cursors

skeqiqevian:
We currently allow users to pattern match via `proc.find`. We will extend that notion to `cursor.find`, enabling pattern matching for cursors.

Also did some miscellaneous code cleanup at the same time. There's a circular import between `API.py` and `API_cursors.py`. I couldn't think of a good way to break the dependency other than merging the two files, anyone have better ideas?

### DIFF
--- a/src/exo/API_cursors.py
+++ b/src/exo/API_cursors.py
@@ -5,17 +5,20 @@ from dataclasses import dataclass
 
 from typing import Optional, List, Any
 
-from . import API
+from . import API  # TODO: remove this circular import
+from .API_types import ExoType
 from .LoopIR import LoopIR
 from .configs import Config
 from .memory import Memory
 
 from . import internal_cursors as C
+from .pattern_match import match_pattern
 from .prelude import Sym
 
 # expose this particular exception as part of the API
 from .internal_cursors import InvalidCursorError
 from .LoopIR_pprint import _print_cursor
+from .LoopIR_scheduling import SchedulingError
 
 
 # --------------------------------------------------------------------------- #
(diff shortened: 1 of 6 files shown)


yamaguchi1024:
Nice. Is this ready for review?

skeqiqevian:
@yamaguchi1024 It should be functional. There are some stylistic choices that are up for debate with regards to code organization.

src/exo/API_cursors.py
```diff
@@ -1005,6 +1011,38 @@ def lift_cursor(impl, proc):
...
+    default_match_no = None if many else 0
+    raw_cursors = match_pattern(
+        scope, pattern, call_depth=1, default_match_no=default_match_no
+    )
+    assert isinstance(raw_cursors, list)
+    cursors = []
+    for c in raw_cursors:
+        c = lift_cursor(c, proc)
+        if isinstance(c, (BlockCursor, ExprListCursor)) and len(c) == 1:
+            c = c[0]
```
yamaguchi1024, line 1037: This line looks a bit suspicious, isn't this saying that you'll return the first statement of the block if `.find` matches with multiple statements? Can you add a test for querying and scoping with block cursors?

skeqiqevian: That's not what this is doing. Each element of `raw_cursor` is a match. This is more of an artifact of how `pattern_match` interacts with the internal `BlockCursor`. When we match a statement `s` in a body or some expr in a list of accesses (e.g. `i` in `x[i, j, i]`), `match_pattern` returns a list (e.g. `[s]` or `[i]`). This code is to essentially unwrap that before returning it to the user.

yamaguchi1024: I see that makes sense

yamaguchi1024 (approved):

PR #545, opened 2023-12-21
Implement remove_control operation

SamirDroubi:
* Remove control and inline the body of an if statement.
* Supported when effects of the statements within the body are only observed when the if condition holds and we don't introduce any out-of-bound accesses.

### DIFF
--- a/src/exo/API_scheduling.py
+++ b/src/exo/API_scheduling.py
@@ -2134,6 +2134,32 @@ def lift_scope(proc, scope_cursor):
     return Procedure(ir, _provenance_eq_Procedure=proc, _forward=fwd)
 
 
+@sched_op([IfCursorA])
+def remove_control(proc, if_cursor):
+    """
+    Eliminate the control introduced by the if statement and replace
+    the if statement by the code within the body.
+
+    This rewrite works only when effects (writes) within the body are only
+    observed when the if condition holds. As a result, if writes happen
+    to any procedure arguments or configuration state, then the rewrite
+    isn't allowed.
+
+    Note: that this is different from when the the condition is always
+    true or false, which is handled by `eliminate_dead_code`.
+
+    rewrite:
+        `if p:`
+        `    s1`
+        -> (assuming effects of `s1` are only observed when `p` holds)
+        `s1`
+    """
+    if_c = if_cursor._impl
+
+    ir, fwd = scheduling.DoRemoveControl(if_c)
+    return Procedure(ir, _provenance_eq_Procedure=proc, _forward=fwd)
+
+
 @sched_op([ForOrIfCursorA])
 def eliminate_dead_code(proc, stmt_cursor):
     """
--- a/src/exo/LoopIR.py
+++ b/src/exo/LoopIR.py
@@ -1088,6 +1088,28 @@ def is_const_zero(e):
     return isinstance(e, LoopIR.Const) and e.val == 0
 
 
+class GetConfigWrites(LoopIR_Do):
+    def __init__(self):
+        self.writes = []
+
+    def do_s(self, s):
+        if isinstance(s, LoopIR.WriteConfig):
+            self.writes.append(s)
+        elif isinstance(s, LoopIR.Call):
+            self.writes += get_config_writes_of_stmts(s.f.body)
+        super().do_s(s)
+
+    # early exit
+    def do_e(self, e):
+        return
+
+
+def get_config_writes_of_stmts(stmts):
+    gwc = GetConfigWrites()
+    gwc.do_stmts(stmts)
+    return gwc.writes
+
+
 class FreeVars(LoopIR_Do):
     def __init__(self, node):
         assert isinstance(node, list)
--- a/src/exo/LoopIR_scheduling.py
+++ b/src/exo/LoopIR_scheduling.py
@@ -12,6 +12,7 @@ from .LoopIR import (
     get_reads_of_expr,
     get_reads_of_stmts,
     get_writes_of_stmts,
+    get_config_writes_of_stmts,
     is_const_zero,
 )
 from .LoopIR_dataflow import LoopIR_Dependencies
@@ -3489,29 +3490,114 @@ class DoSimplify(Cursor_Rewrite):
             raise NotImplementedError(f"bad case {type(s)}")
 
 
-def DoEliminateIfDeadBranch(if_cursor):
+def _get_alloc(stmt_cursor, name):
+    for arg in stmt_cursor.root()._child_block("args"):
+        if arg._node.name == name:
+            return arg
+    while True:
+        node = stmt_cursor._node
+        assert not isinstance(node, LoopIR.proc), "Unreachable, must find allocation"
+        if isinstance(node, LoopIR.Alloc) and node.name == name:
+            return stmt_cursor
+        elif stmt_cursor.get_index() == 0:
+            stmt_cursor = stmt_cursor.parent()
+        else:
+            stmt_cursor = stmt_cursor.prev()
+
+
+def _Check_BlockAccessesBounds(ir, block):
+    nodes = [s._node for s in block]
+    accesses = get_reads_of_stmts(nodes) + get_writes_of_stmts(nodes)
+    accesses = set(name for name, typ in accesses if typ.is_numeric())
+    for access in accesses:
+        alloc = _get_alloc(block[-1], access)
+        alloc_within_block = any(stmt.is_ancestor_of(alloc) for stmt in block)
+        if not alloc_within_block:
+            Check_Bounds(ir, alloc._node, nodes)
+        else:
+            stmts_after_alloc = [
+                stmt._node for stmt in alloc.as_block().expand(delta_lo=-1)
+            ]
+            Check_Bounds(ir, alloc._node, stmts_after_alloc)
+
+
+def _Check_IfEffectsObservedUnderCond(ir, if_cursor):
+    assert not if_cursor._node.orelse
+
+    nodes = [s._node for s in if_cursor.body()]
+
+    # Check for effects on any configuration state
+    if get_config_writes_of_stmts(nodes):
+        raise SchedulingError(
+            f"If statement body has effects on a configuration state; "
+            + "effects can be observed when the condition doesn't hold (from the caller)"
+        )
+
+    # Check for effects on any data arguments
+    writes = get_writes_of_stmts(nodes)
+    writes = set(name for name, typ in writes if typ.is_numeric())
+    for write in writes:
+        alloc = _get_alloc(if_cursor.body()[-1], write)
+        if isinstance(alloc._node, LoopIR.fnarg):
+            raise SchedulingError(
+                f"If statement body has effects on argument {alloc._node.name} of the procedure; "
+                + "effects can be observed when the condition doesn't hold (from the caller)"
+            )
+        # TODO: We should be able to relax this condition
+        if alloc.parent() != if_cursor.parent() and not if_cursor.is_ancestor_of(alloc):
+            raise SchedulingError(
+                "If statement must be in the same scope of the allocations it affects"
+            )
+
+    # Check that effects on local data variables are only observed when the if condition holds
+    cond = if_cursor._node.cond
+    true_node = LoopIR.Const(True, T.bool, cond.srcinfo)
+
+    def check(block):
+        for stmt in block:
+            node = stmt._node
+            if isinstance(node, LoopIR.For):
+                check(stmt.body())
+            elif isinstance(node, LoopIR.If):
+                check(stmt.body())
+                check(stmt.orelse())
+            else:
+                accesses = get_reads_of_stmts([node]) + get_writes_of_stmts([node])
+                accesses = set(name for name, _ in accesses)
+                if len(writes & accesses):
+                    try:
+                        Check_ExprEqvInContext(ir, cond, [stmt._node], true_node)
+                    except SchedulingError:
+                        raise SchedulingError(
+                            "Writes within the if statement are observed when the condition doesn't hold"
+                        )
+
+    block_after = if_cursor.as_block().expand(delta_lo=-1)
+    check(block_after)
+
+
+def DoRemoveControl(if_cursor):
     if_stmt = if_cursor._node
 
     assert isinstance(if_stmt, LoopIR.If)
 
-    ir, fwd = if_cursor.get_root(), lambda x: x
+    """
+    TODO: The difficult part with this is making sure that effect of the if branch don't affect the else branch.
+    We can handle this by splitting the if/else statement into two if statements with no else branch. 
+    Then, removing the control from both. This splitting rewrite could itself be its own operation.
+    """
+    if if_stmt.orelse:
+        raise SchedulingError("Cannot have an else branch")
 
-    try:
-        cond_node = LoopIR.Const(True, T.bool, if_stmt.srcinfo)
-        Check_ExprEqvInContext(ir, if_stmt.cond, [if_stmt], cond_node)
-        cond = True
-    except SchedulingError:
-        try:
-            cond_node = LoopIR.Const(False, T.bool, if_stmt.srcinfo)
-            Check_ExprEqvInContext(ir, if_stmt.cond, [if_stmt], cond_node)
-            cond = False
-        except SchedulingError:
-            raise SchedulingError("If condition isn't always True or always False")
+    _Check_IfEffectsObservedUnderCond(if_cursor.get_root(), if_cursor)
 
-    body = if_cursor.body() if cond else if_cursor.orelse()
+    body = if_cursor.body()
     ir, fwd = body._move(if_cursor.after())
     ir, fwd_del = fwd(if_cursor)._delete()
     fwd = _compose(fwd_del, fwd)
+
+    stmts = [fwd(s) for s in body]
+    _Check_BlockAccessesBounds(ir, stmts)
 
     return ir, fwd
 
@@ -4227,7 +4313,7 @@ __all__ = [
     "DoFissionLoops",
     "DoBoundAndGuard",
     "DoDeletePass",
-    "DoEliminateDeadCode",
+    "DoRemoveControl" "DoEliminateDeadCode",
     "DoAddUnsafeGuard",
     "DoStageWindow",
     "DoBoundAlloc",
--- a/src/exo/new_eff.py
+++ b/src/exo/new_eff.py
@@ -731,7 +731,6 @@ def _lsstr(ls, prec=0):
         typ = f":{ls.type}" if ls.type else ""
         return f"{{({ls.name},{coords}{typ})}}"
     elif isinstance(ls, LS.WholeBuf):
-        coords = ",".join([str(a) for a in ls.coords])
         return f"{{{ls.name}|{ls.ndim}}}"
     elif isinstance(ls, (LS.Union, LS.Isct, LS.Diff)):
         if isinstance(ls, LS.Union):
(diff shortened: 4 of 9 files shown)


yamaguchi1024:
Is this the same or more general version of assert_if? Btw what happened to assert_if?

SamirDroubi:
> Is this the same or more general version of assert_if? Btw what happened to assert_if?

No, `assert_if` used to reason whether the condition is alawys true or alawys false. This reasons about whether it is okay to remove the if control and execute the body regardless of whether the condition is always true or flase. `assert_if` was doing dead code elimination and so it is behavior is now a part of `eleminate_dead_code`. This is not more general form of `assert_if`.

yamaguchi1024:
Ah gotcha. Is remove_if part of eleminate_dead_code too?

SamirDroubi:
> Ah gotcha. Is remove_if part of eleminate_dead_code too?

Originally, `assert_if` has the API: `assert_if(proc, if, cond_value)`. Then, we automated choosing the `cond_value` and renamed it to `remove_if(proc, if)`. Then, we realized it is part of a more general concept which is dead code elimination. The API only exposes one primitive called `eliminate_dead_code` which dispatches to mutliple internal implementation: one for if statements (what used to be `remove_if`) and one for loops; there are more cases but we don't have implementations for them for now.

yamaguchi1024:
Aren't eliminate_dead_code and remove_control inherently the same though? remove_control will rewrite
```
if cond:
  s1
```
to
```
s1
```
if `cond` is always true, but this is the same as rewriting
```
if cond:
  s1
else:
  pass
```
into
```
s1
```
by eliminate_dead_code. The only difference is whether or not you have explicit else branch.

SamirDroubi:
The rewrite is sortof the same, but the reason why you can do the rewrite is different.

In eliminate_dead_code, you are saying one of the branches is dead and will never be executed. In remove_control, you are saying the else branch isn't neccesarily dead, but removing the control still preserves equivelance.

Maybe there need to be a better name to signify the difference. 

Also, `eliminate_dead_code` can rewrite:
```
if cond:
     s1
else:
    s2
---->
s2
```
While `remove_control` would rewrite:
```
if cond:
     s1
else:
    s2
---->
s1
s2
```
Currently it only works when s2=pass, but I can generalize it. I didn't implement it for now because it is not an important use case.

yamaguchi1024:
I'm confused... This rewrite is strictly incorrect, right? 
```
if cond:
  s1
else:
  s2
---->
s1
s2
```

SamirDroubi:
The simplest example where this is useful is something like the following:
```
a: f32
a = 0.0
if n > 0:
    a = 1.0
for i in seq(0, n):
   b += a
```
Here you can remove the if statement because, every time the value of `a` is accessed, the condition `n > 0` holds.

SamirDroubi:
This is an example where that rewrite is correct:
```
a: f32
b: f32
if n > 5:
   a = 1.0
else:
   b = 1.0
for i in seq(6, n):
    c += a
for i in seq(0, 6):
    c += b
```
Which is equivelant to:
```
a: f32
b: f32
a = 1.0
b = 1.0
for i in seq(6, n):
    c += a
for i in seq(0, 6):
    c += b
```

yamaguchi1024:
We absolutely shouldn't eliminate if branch based on the effects after the statement...

SamirDroubi:
Why not?

yamaguchi1024:
Making rewrites depends on the global effects won't generalize, it will look like compile-time runtime very quickly. Ultimately, it will lead to sth like rewriting
```
def fib(...):
  assert n == 10
  .... # actually compute fib
```
into
```
def fib(...):
  assert n == 10
  return 55
```

SamirDroubi:
I don't fully understand what the first statement means. The two programs you wrote are equivelant, I don't see the problem.

SamirDroubi:
If a rewrite turns a program `P` into `P'` and they both have the same effects on all the arguments and global state. Why should that rewrite be illegal?

yamaguchi1024:
I'm not saying it's illegal, but I think primitives that require overly complex analysis have a danger of becoming brittle, like a giant pattern match. I think we should stick to local rewrite and local analysis as much as possible, unless when there is no alternative way to rewrite. What was the motivation for introducing this operation? Can you show me the application code that you wanted to schedule?

SamirDroubi:
I don't the analysis is brittle right now, and it is fairly simple. The use cases are the examples I added as tests.

This is another example
```
for j in seq(0, (n + 7) / 8):
      a_temp: f32
       if i < n:
            a_temp = a[i][k]
       b_temp: f32
       if i < n:
          b_temp = b[k][j]
       if i < n:
           c[i, j] += a_temp * b_temp
```
We can use the current operation to remove the if around the read to `a` to get:
```
for j in seq(0, (n + 7) / 8):
      a_temp: f32
      a_temp = a[i][k]
      b_temp: f32
      if i < n:
         b_temp = b[k][j]
      if i < n:
          c[i, j] += a_temp * b_temp
```
Then, the same analysis (not rewrite) can be used to turn the load to `b` into a masked load. 
```
for j in seq(0, (n + 7) / 8):
      a_temp: f32
      a_temp = a[i][k]
      b_temp: f32
      if i < n:
         b_temp = b[k][j]
      else:
         b_temp = 0.0
      if i < n:
          c[i, j] += a_temp * b_temp
```

yamaguchi1024:
What is the bound condition for `i < n` ? Is it always true? If so, eliminate_dead_code should suffice, and if not the guard around `a_temp = a[i, k]` shouldn't be removed since it will be read out of bounds.

SamirDroubi:
Sorry, there was a type above the condition should be `j < n`. It is fixed now

yamaguchi1024:
So `j < n` is always true in the `j` loop if `n > 1`, and if there's `assert n > 1` in the proc, rewriting
```
if j < n:
  a_temp = a[i, k]
--->
a_temp = a[i, k]
```
should be always correct and it should be handled by eliminate_dead_code IMO. If you cannot assert `n > 1`, this rewrite isn't correct.

SamirDroubi:
The example was missing something again, I just fixed it.

yamaguchi1024:
I understood that you want to remove the guard of
```
if jo * 8 + ji < n:
  a_temp = a[i, k]
```
because the body is idempotent w.r.t the `j` loop, but I think in this case you don't want to introduce the guard `if jo * 8 + ji < n:` in the first place. Why was this condition introduced?

SamirDroubi:
`divide_loop` with the `guard` option.

yamaguchi1024:
I had similar situation for gemmini matmul, and I did something like the following rewrite
```
for jo in seq(0, (n + 7) / 8):
   for ji in seq(0, 8):
      a_temp: f32
       if jo * 8 + ji < n:
            a_temp = a[i, k]
       ...
----> lift_alloc
a_temp: f32
for jo in seq(0, (n + 7) / 8):
   for ji in seq(0, 8):
       if jo * 8 + ji < n:
            a_temp = a[i, k]
       ...
----> fission
a_temp: f32
for jo in seq(0, (n + 7) / 8):
   for ji in seq(0, 8):
       if jo * 8 + ji < n:
            a_temp = a[i, k]
for jo in seq(0, (n + 7) / 8):
   for ji in seq(0, 8):
       ...
----> remove_if and remove_loop
a_temp: f32
a_temp = a[i, k]
for jo in seq(0, (n + 7) / 8):
   for ji in seq(0, 8):
       ...
----> add_loop
a_temp: f32
for jo in seq(0, (n + 7) / 8):
   for ji in seq(0, 8):
         a_temp = a[i, k]
for jo in seq(0, (n + 7) / 8):
   for ji in seq(0, 8):
       ...
----> fuse
a_temp: f32
for jo in seq(0, (n + 7) / 8):
   for ji in seq(0, 8):
       a_temp = a[i, k]
       ...
```

yamaguchi1024:
I think the same result can be achieved by composing existing primitives, and I think we should implement a standard library of this rewrite sequence instead of creating a new primitive.

SamirDroubi:
The step where you remove the loops wouldn't work becuase the variables `ji` and `jo` aren't free in the body.

This also still wouldn't handle the rewrite of inserting an else branch that writes a zero when you reading out of bounds (like a masked load would do).

yamaguchi1024:
It'll be even easier if you fission after `a_temp = ...` before divide_loop. And then add_loop

SamirDroubi:
```
    @proc
    def foo(M: size, N: size, K: size, a: f32[M, K]):
        a_temp: f32
        for k in seq(0, K):
            for i in seq(0, M):
                for jo in seq(0, (N + 7) / 8):
                    for ji in seq(0, 8):
                        if jo * 8 + ji < N:
                            a_temp = a[i, k]
    foo = remove_loop(foo, "ji")
 exo.new_eff.SchedulingError: <<<unknown directive>>>: Cannot remove loop, ji is not free in the loop body
```

yamaguchi1024:
You should remove_loop before dividing

SamirDroubi:
I will try this later. I don't see how you would be able to generate masked loads with this approach though.

gilbo:
I haven’t followed all the details here, but I want to back up Yuka’s initial instinct. There is no way thatif c:  s1else:  s2is generally safe to rewrite tos1; s2This is only achievable when s1 and s2 have “magically” disjoint downstream consequences.On Dec 20, 2023, at 10:01 PM, Samir Droubi ***@***.***> wrote:﻿
I will try this later. I don't see how you would be able to generate masked loads with this approach though.

SamirDroubi:
To be clear, rewrite `if c: s1else: s2` into the general case doesn't really matter. In practice, what really matters is the case where `s2=pass`. I was just using that general case to signify the difference between this operation and dead code elimination (because the rewrite could look the same when `s2=pass`).

-----
Let me try to explain this from the top.
Consider the following programs:

_Example1:_
```
def scal(n: size, alpha: f32 @ DRAM, x: [f32][n] @ DRAM):
    assert stride(x, 0) == 1
    alpha_: f32 @ DRAM
    alpha_ = alpha
    for io in seq(0, n / 8):
        reg: f32[8] @ AVX2
        mm256_broadcast_ss_scalar(reg[0:8], alpha_)
        reg_1: f32[8] @ AVX2
        mm256_loadu_ps(reg_1[0:8], x[8 * io + 0:8 * io + 8])
        reg_2: f32[8] @ AVX2
        mm256_mul_ps(reg_2[0:8], reg[0:8], reg_1[0:8])
        mm256_storeu_ps(x[8 * io + 0:8 * io + 8], reg_2[0:8])
    for ii in seq(0, n % 8):
        x[ii + n / 8 * 8] = alpha_ * x[ii + n / 8 * 8]
```
A natural thing to want to do is hoist the broadcast:
```
scal = lift_alloc(scal, "reg")
scal = fission(scal, scal.find_loop("io").body()[0].after())
scal = remove_loop(scal, scal.find_loop("io"))
---------->
def scal(n: size, alpha: f32 @ DRAM, x: [f32][n] @ DRAM):
    assert stride(x, 0) == 1
    alpha_: f32 @ DRAM
    alpha_ = alpha
    reg: f32[8] @ AVX2
    if n / 8 > 0:
        mm256_broadcast_ss_scalar(reg[0:8], alpha_)
    for io in seq(0, n / 8):
        reg_1: f32[8] @ AVX2
        mm256_loadu_ps(reg_1[0:8], x[8 * io + 0:8 * io + 8])
        reg_2: f32[8] @ AVX2
        mm256_mul_ps(reg_2[0:8], reg[0:8], reg_1[0:8])
        mm256_storeu_ps(x[8 * io + 0:8 * io + 8], reg_2[0:8])
    for ii in seq(0, n % 8):
        x[ii + n / 8 * 8] = alpha_ * x[ii + n / 8 * 8]
```
I have successfuly hoisted the braodcast, but `remove_if` add an if statement around it because it is the loop that was removed may not run at least once. Now, let's say I want to remove this if statement. First question is why would I wanna be able to do that?
1. Removing it doesn't cause any out of bounds accesses.
2. Everytime `reg` is read later one (which is within the first loop), the condition `n / 8 > 0` holds. In other words, if I get to a point where I have to read/write `reg` after the if statement, then the if statement must have been true when I executed it.
```
scal = remove_control(scal, "if _ : _")
---------->
def scal(n: size, alpha: f32 @ DRAM, x: [f32][n] @ DRAM):
    assert stride(x, 0) == 1
    alpha_: f32 @ DRAM
    alpha_ = alpha
    reg: f32[8] @ AVX2
    mm256_broadcast_ss_scalar(reg[0:8], alpha_)
    for io in seq(0, n / 8):
        reg_1: f32[8] @ AVX2
        mm256_loadu_ps(reg_1[0:8], x[8 * io + 0:8 * io + 8])
        reg_2: f32[8] @ AVX2
        mm256_mul_ps(reg_2[0:8], reg[0:8], reg_1[0:8])
        mm256_storeu_ps(x[8 * io + 0:8 * io + 8], reg_2[0:8])
    for ii in seq(0, n % 8):
        x[ii + n / 8 * 8] = alpha_ * x[ii + n / 8 * 8]
```

_example 2:_
```
def gemm(M: size, N: size, K: size,
                                A: [f32][M, K] @ DRAM, B: [f32][K, N] @ DRAM,
                                C: [f32][M, N] @ DRAM):
    assert stride(A, 1) == 1
    assert stride(B, 1) == 1
    assert stride(C, 1) == 1
    assert K <= 192
    assert M <= 4096
    assert N <= 3072
    for k in seq(0, K):
        for i in seq(0, M):
            for jo in seq(0, (N + 7) / 8):
                for ji in seq(0, 8):
                    A_temp: f32
                    B_temp: f32
                    C_temp: f32 @ DRAM
                    if 8 * jo + ji < N:
                        A_temp = A[i, k]
                    if 8 * jo + ji < N:
                        B_temp = B[k, 8 * jo + ji]
                    if 8 * jo + ji < N:
                        C_temp = C[i, jo * 8 + ji]
                    if 8 * jo + ji < N:
                        C_temp += A_temp * B_temp
                    if 8 * jo + ji < N:
                        C[i, jo * 8 + ji] = C_temp
```
I should be able to remove the guard around the load from `A_temp`. I should be able to do so because:
1. `A[i, k]` wouldn't be accessed out of bounds.
2. `A_temp` is only accessed when the condition `8 * jo + ji < N` holds.
```
gemm = remove_control(gemm, "if _ : _")
---->

def gemm(M: size, N: size, K: size,
                                A: [f32][M, K] @ DRAM, B: [f32][K, N] @ DRAM,
                                C: [f32][M, N] @ DRAM):
    assert stride(A, 1) == 1
    assert stride(B, 1) == 1
    assert stride(C, 1) == 1
    assert K <= 192
    assert M <= 4096
    assert N <= 3072
    for k in seq(0, K):
        for i in seq(0, M):
            for jo in seq(0, (N + 7) / 8):
                for ji in seq(0, 8):
                    A_temp: f32
                    B_temp: f32
                    C_temp: f32 @ DRAM
                    A_temp = A[i, k]
                    if 8 * jo + ji < N:
                        B_temp = B[k, 8 * jo + ji]
                    if 8 * jo + ji < N:
                        C_temp = C[i, jo * 8 + ji]
                    if 8 * jo + ji < N:
                        C_temp += A_temp * B_temp
                    if 8 * jo + ji < N:
                        C[i, jo * 8 + ji] = C_temp
```
So, the safety conditions for removing an if statement and replacing it with its body (still assuming there is no else branch):
1. It doesn't result in any out-if-bound access.
2. The effects of the body are only observed when the condition of the if statement holds.

Now, to handle the case where you have an `else` branch (which doesn't really matter in-practice): you rewrite the if/else into two ifs, then remove each one separtely according to the two conditions above.
```
if c:
   s1
else:
    s2
----->
if c:
    s1
if !c:
    s2
---->
s1
if !c:
   s2
---->
s1
s2
```

Finally, how can we handle generating masked loads using this analysis (not the operation). Let's consider the gemm example again:
```
def gemm(M: size, N: size, K: size,
                                A: [f32][M, K] @ DRAM, B: [f32][K, N] @ DRAM,
                                C: [f32][M, N] @ DRAM):
    assert stride(A, 1) == 1
    assert stride(B, 1) == 1
    assert stride(C, 1) == 1
    assert K <= 192
    assert M <= 4096
    assert N <= 3072
    for k in seq(0, K):
        for i in seq(0, M):
            for jo in seq(0, (N + 7) / 8):
                for ji in seq(0, 8):
                    A_temp: f32
                    B_temp: f32
                    C_temp: f32 @ DRAM
                    A_temp = A[i, k]
                    if 8 * jo + ji < N:
                        B_temp = B[k, 8 * jo + ji]
                    if 8 * jo + ji < N:
                        C_temp = C[i, jo * 8 + ji]
                    if 8 * jo + ji < N:
                        C_temp += A_temp * B_temp
                    if 8 * jo + ji < N:
                        C[i, jo * 8 + ji] = C_temp
```
I would like to transform it into:
```
def gemm(M: size, N: size, K: size,
                                A: [f32][M, K] @ DRAM, B: [f32][K, N] @ DRAM,
                                C: [f32][M, N] @ DRAM):
    assert stride(A, 1) == 1
    assert stride(B, 1) == 1
    assert stride(C, 1) == 1
    assert K <= 192
    assert M <= 4096
    assert N <= 3072
    for k in seq(0, K):
        for i in seq(0, M):
            for jo in seq(0, (N + 7) / 8):
                for ji in seq(0, 8):
                    A_temp: f32
                    B_temp: f32
                    C_temp: f32 @ DRAM
                    A_temp = A[i, k]
                    if 8 * jo + ji < N:
                        B_temp = B[k, 8 * jo + ji]
                    else:
                        B_temp = 0.0
                    if 8 * jo + ji < N:
                        C_temp = C[i, jo * 8 + ji]
                    if 8 * jo + ji < N:
                        C_temp += A_temp * B_temp
                    if 8 * jo + ji < N:
                        C[i, jo * 8 + ji] = C_temp
```
Which is an safe because:
1. I did't introduce any out-of-bound acccesses in the else branch.
2. Two the effects on `B_temp` are only observed when the condition `8 * jo + ji < N` holds.
i.e. the same two safety conditions (analysis) allows me to generate masked loads.

So, now I can implement an operation of the following:
```
mirror_if_branch(proc, if, List(NewExprA))
```
and use it in the following way:
```
gemm = mirro_if_branch(proc, "if _:_", 0.0)
```

yamaguchi1024:
Example 1: We know `n > 0` already from the `size` type, so `n / 8 > 0` is always true regardless of the statements afterward. So I think eliminate_dead_coed should work.

Example 2: I think you want to lift and hoist A_temp before dividing the `j` loop. That will free you from having to remove `8 * jo + ji < N` from A_temp

SamirDroubi:
In example 1, `n / 8` isn't always greater than 0; the integer division is equal to zero when `n = 1`.

yamaguchi1024:
Lol yeah that's true, can we leave the guard as is or assert n > 8?

SamirDroubi:
asserting `n > 8` isn't correct. You can leave the guard, but that's not the code you would want to generate.

yamaguchi1024:
If n <= 8 you'll end up doing redundant broadcast so you probably don't want to remove the guard anyways

SamirDroubi:
I simplified the example, but if you have a vectorized tail loop, then you would want to share the broadcast.

SamirDroubi:
You might be right about example 2, but I think this only works because the loop bound is `N`. In the general case, you are relying on `remove_loop` which could potentially insert extra guards because the loop may not run and you won't be able to follow the same flow you are talking about. In the gemm example, replace the loop upper bound with `N - 1` and I don't think what you are describing would work.

yamaguchi1024:
>  replace the loop upper bound with N - 1 and I don't think what you are describing would work.

It will still work as long as the body statement that you are trying to remove the loop on (in this case `A_temp = A[i, k]`) does not depend on `j`

SamirDroubi:
I don't see how I can generate a vector broadcast if I hoist the read to `alpha` before dividing the loop.

yamaguchi1024:
I thought you were talking about hoisting vector broadcasts for kernels with multiple dimensions and multiple tail cases that are more complex than one-dimensional scal. For the broadcasting dimension you have to divide loop first but for other dimensions you can hoist the broadcast before generating the tail cases, right?

SamirDroubi:
I think the same issue shows up regardless.

yamaguchi1024:
@SamirDroubi is it okay to close this?

PR #546, opened 2023-12-29
Forwarding blocks for edit functions, allow replace to forward when it is 1-to-1

skeqiqevian:
Currently, forwarding doesn't work for any blocks. However, there are many blocks which lie outside of edit scopes where forwarding behavior should be very natural. We try to support such behavior for each of the five edit functions. 

# Insertion
All blocks will forward. Block which contain the insertion gap should forward to a block which includes the inserted lines of code. 

# Deletion
- Blocks entirely containing the deletion block will forward. 
- Blocks inside the deleted block will not forward.
- Blocks partially containing the deleted block will not forward.

# Replace

## Replacing a node with node
In this case, the old node will forward to the new node. All blocks forward. 

## Replacing a node/block with a block
Forwarding behavior is identical to deletion of a block.

# Wrap
- Blocks entirely containing the wrapped block will forward.
- Blocks inside the wrapped block will forward
- Blocks partially containing the wrapped block will not forward.

# Move
All blocks forward except those which partially contain the moved block.


SamirDroubi:
I have been trying to read through the code since yesterday, but it will take me a bit of time to follow all the details since I am not very familiar with this part of the codebase. One note that I had is that it seems like this change will propagate to the API. I think it would be nice to have at least one test for all these different cases showing where a block will forward after one or more scheduling operations. This would also exercise some interactions between forwarding strategies of different cursor edits (on different type types of cursors: nodes/blocks/gaps).

yamaguchi1024:
Same here, I asked Kevin to walk me through the PR today, you're welcome to join

yamaguchi1024 (approved):
@skeqiqevian feel free to merge it whenever you feel is ready

Issue #557, opened 2024-01-23
Check_ReorderLoop fails when reduction is not in reduction syntax


SamirDroubi:
The following program returns `unknown` result on z3 4.12.4.0 (and unsat result on z3 4.12.5.0):
```
    @proc
    def foo(n: size, x: [f32][n] @ DRAM, y: [f32][n] @ DRAM,
                      result: f32 @ DRAM):
        assert stride(x, 0) == 1
        assert stride(y, 0) == 1
        result_: f32 @ DRAM
        result_ = 0.0
        var0: f32[8] @ AVX2
        for ii in seq(0, 8):
            var0[ii] = 0.0
        for ioo in seq(0, ((7 + n) / 8 - 1) / 4):
            for ioi in seq(0, 4):
                var1: f32[8] @ AVX2
                var2: f32[8] @ AVX2
                for ii in seq(0, 8):
                    var1[ii] = x[ii + 8 * (4 * ioo + ioi)]
                for ii in seq(0, 8):
                    var2[ii] = y[ii + 8 * (4 * ioo + ioi)]
                for ii in seq(0, 8):
                    var0[ii] = var1[ii] * var2[ii] + var0[ii]
        for ii in seq(0, 8):
            result_ += var0[ii]
        result = result_
``` 
However, switching the following assignment to a reduction:
```
var0[ii] = var1[ii] * var2[ii] + var0[ii]
--->
var0[ii] += var1[ii] * var2[ii]
```
Let's it pass. Is this an expected behavior or a bug?

-------------------------------------------------------------------------------------------------------
Here is the debug info when the statement is in assignment form:
```
assume  M(stride(x,0) == 1 ∧ stride(y,0) == 1 ∧ n > 0)
to verify
[...]

PR #558, opened 2024-01-24
Implement fold_into_reduce and reassociate_expr

SamirDroubi:
This PR implements two operations:
1. fold_into_reduce: folds an assignment into a reduction format
2. reassociate_expr: changes the associativity of addition and multiplication. This can be used to help with (1).

The primary reason behind this PR is being able to put assignments into reduction format which some analysis depends on (e.g. #557).


src/exo/API_scheduling.py
```diff
@@ -902,6 +902,35 @@ def commute_expr(proc, expr_cursors):
...
+            ->
+        (a + b) + c
+    """
+    expr = expr._impl
+    if not isinstance(expr._node, LoopIR.BinOp) or (
+        expr._node.op != "+" and expr._node.op != "*"
+    ):
+        raise TypeError(f"Only '+' or '*' can be reassociated, got {expr._node.op}")
+    if (
+        not isinstance(expr._node.rhs, LoopIR.BinOp)
```
yamaguchi1024, line 924: Wouldn't it make sense to reassociate LHS `(a + b) + c` as well?

SamirDroubi: Reassociate `(a + b) +c` to `a + (b + c)` can be done this way:
`(a + b) +c` --commute--> `c + (a + b)` --commute--> `c + (b + a)` --reassociate--> `(c + b) + a` --commute--> ` a + (c + b)` --commute--> ` a + (b + c)`

We can provide such wrapper in the standard library.

src/exo/LoopIR_scheduling.py
```diff
@@ -492,6 +492,32 @@ def DoMergeWrites(c1, c2):
...
+        idx = f"[{','.join([str(idx) for idx in node.idx])}]" if node.idx else ""
+        return f"{node.name}{idx}"
+
+    assign_s = assign._node
+
+    if not isinstance(assign_s.rhs, LoopIR.BinOp) or assign_s.rhs.op != "+":
+        raise SchedulingError("The rhs of the assignment must be an add.")
+    if not isinstance(assign_s.rhs.rhs, LoopIR.Read) or access_to_str(
+        assign_s
+    ) != access_to_str(assign_s.rhs.rhs):
```
yamaguchi1024, line 506: RHS is being hardcoded here, this will not be able to fold `a = a + 1.0`, right? I understand that users can commute to have `a` on the RHS that's confusing IMO

SamirDroubi: I originally wrote it to handle all sorts of cases. But then I realized there is a billion cases to handle. So, I decided to make it as concise and deterministic as possible in accordance with the primitives design.

We can build automations that can handle this arbitrary cases in the standard library: handle arbitrary reassociation and commute to put the expression in the right shape and then fold it.

src/exo/API_scheduling.py
yamaguchi1024: I guess that'll be ambiguos for some cases e.g., `(a + b) + (c + d)`, it is not clear how it will be associated if LHS is also supported.

src/exo/LoopIR_scheduling.py
yamaguchi1024: How about communicating that to users via primitive names? `reassociate_expr` and `fold_into_reduce` look like they are general primitives that work on arbitrary expressions. Maybe rename them `reassociate_rhs_expr` and `fold_rhs_into_reduce` or something like that?

SamirDroubi: That's fair I think. I have been thinking about the same thing actually.

src/exo/API_scheduling.py
SamirDroubi: I don't follow your statement. Can you clarify?

src/exo/LoopIR_scheduling.py
SamirDroubi: I think the operation would be called `left_reassociate_expr` since this format is (a + b) + c means that addition has left associativity.

SamirDroubi: I am not actually sure about this name `fold_rhs_into_reduce`. Because we pass a cursor to an assign statement. So, is the rhs for the assignment or the rhs of the expression that is on the rhs of the assignment.

src/exo/API_scheduling.py
yamaguchi1024: I was trying to say that hardcoding the direction of reassociation (as you do rn) is correct instead of supporting LHS reassociation in the same primitive, because reassociating `(a + b) + (c + d)` might result in `(a) + (b + (c + d))` or `((a + b) + c) + (d)` depending on the direction, which is ambiguous.

src/exo/LoopIR_scheduling.py
SamirDroubi: One option is to simply support both the LHS and the RHS of this operation.

SamirDroubi: I think this name `fold_into_reduce` might be fine. `a = a + 1` -> `a += 1` wouldn't really be folding. While `a = 1 + a` -> `a += 1` would be folding. At least if you think about the physical movement of folding.

yamaguchi1024: It should be either "right most" or "left most" expressions, right?
Because the expression can be sth like `a = (3.0 + (2.0 + a))` in which case you'll want to traverse down to the right most expression

SamirDroubi: Hmmm I might have been wrong about the canonical way in what direction to go:
https://www.ibm.com/docs/en/i/7.2?topic=operators-compound-assignment

SamirDroubi: In the case you wrote about, you would reassociate first, then fold.

SamirDroubi: I think I should change the rewrite to be `a = a + ...` -> `a += ...`. There is still a question of what to call it.

SamirDroubi: There must be a canonical name for this rewrite but I can't find it anywhere.

yamaguchi1024:
`left_reassociate_expr` looks good because it doesn't care about the content of A, B, and C as long as they are in this AST form:
PXL_20240124_203500468

I still think it's confusing for `fold_into_reduce` to accept the left figure while not accepting the right figure, even though they both look like `a = a + b + ...` in the frontend syntax.
PXL_20240124_203720814

SamirDroubi:
They look the same because we simplify the parenthesis. My experience with languages that implements expression rewrites (like Coq) is that you just have to pay attention to these things when using the primitives. Then, you build hammer automations on top of it.

yamaguchi1024:
I agree that we should build automation in the standard library, but at least the primitive definition should be consistent with what it is actually doing. The current documentation writes
```
    rewrite:
        a = a + ...
            ->
        a += ...
```
which doesn't accurately describe the acceptable program. You could change the documentation to describe the left figure above, but I think it will be hard to write the documentation with syntax that is opaque to users (binop AST).

SamirDroubi:
We can change it to:
```
    rewrite:
        a = a + (...)
            ->
        a += ...
```
We also need to a better name that signifies what it does.

yamaguchi1024:
Yeah I guess `a = a + (...)` is consistent. Also, folding reduction is necessary for circumventing the reorder_loop bug, right? So hopefully this won't be an issue when we have better analysis

yamaguchi1024 (approved):

Issue #410, opened 2023-04-13
Simplifying loops with one statement in body that is an if


SamirDroubi:
This was brought up when handling banded storage in BLAS level 2
```
@proc
def tbmv_row_major_Upper_NonTrans_template_with_if(n: size, k: size, x: [R][n], A: [R][n, k + 1], Diag: size):
    assert stride(A, 1) == 1
    assert k <= n - 1
    
    for i in seq(0, n):
        dot: R
        dot = 0.0
        
        if Diag == 0:
            dot = x[i] * A[i, 0]
        else:
            dot = x[i]

        for j in seq(0, k):
            if i + j + 1 < n:
               dot += A[i, j + 1] * x[i + j + 1]
            
        x[i] = dot
```
We should be able to the outer loop with `n-k`
Then, the if statement in the first nested loops will be redundant: `i < n - k, j < k` -> `i + j + 1 < (n - k - 1) + (k - 1) + 1 = n - 1 < n` (this currently can be done with `assert_if`)

And, the inner loop in the second nested loop will be bounded by the conditional in the loop if statement rather the loop hi:
`i < k, j < k` ->  `n - k + i + j + 1 < n` -> `- k + i + j + 1 < 0` -> `j < k - i - 1 < k`

So, we get:
```
@proc
def tbmv_row_major_Upper_NonTrans_template(n: size, k: size, x: [R][n], A: [R][n, k + 1], Diag: size):
    assert stride(A, 1) == 1
    assert k <= n - 1
    
    for i in seq(0, n - k):
        dot: R
        dot = 0.0
        
        if Diag == 0:
            dot = x[i] * A[i, 0]
        else:
            dot = x[i]

        for j in seq(0, k):
            if i + j + 1 < n:
               dot += A[i, j + 1] * x[i + j + 1]
            
        x[i] = dot
        
    for i in seq(0, k):
        dot: R
        dot = 0.0
        
        if Diag == 0:
            dot = x[n - k + i] * A[n - k + i, 0]
        else:
            dot = x[n - k + i]

        for j in seq(0, k - i - 1):
            dot += A[n - k + i, j + 1] * x[n - k + i + j + 1]
            
        x[n - k + i] = dot
```

When we have a statement of the form
```
for i in seq(0, hi):
     if e + ci < e' + c'i:
        B;
```
then, we know that the inner if is equivelant:
```
i < (e' - e)/(c - c')
```
We should be able to simplify the loop to:
```
[...]

PR #561, opened 2024-01-27
Implement bound_loop_by_if

SamirDroubi:
* Addresses #410. This PR implements a primitive that does most of the work for simplifying the case described in the issue. There are other mathematical rewrites that are necessary to get the full simplification in the general case. However, those can be implemented as separate primitives. Then, we can build automations that can handle the general case.

### DIFF
--- a/src/exo/API_scheduling.py
+++ b/src/exo/API_scheduling.py
@@ -2203,6 +2203,30 @@ def eliminate_dead_code(proc, stmt_cursor):
     return Procedure(ir, _provenance_eq_Procedure=proc, _forward=fwd)
 
 
+@sched_op([ForCursorA])
+def bound_loop_by_if(proc, loop):
+    """
+    Bound the loop by the inner if statement.
+
+    Requires:
+        - The inner if to be the only statement in the loop body.
+        - The condition of the if statement emposes a tigher
+            iteration space than that of the loop.
+        - Loop `lo` to be `0`.
+        - The if condition to be of the form `iter < e`.
+
+    Rewrite:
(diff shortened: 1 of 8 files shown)


SamirDroubi:
There is a lot of checks in the implementation of the primitive that resembles a pattern_match. In the process of trying to capture those checks as a pattern, I discovered various issues in the way we currently do pattern matching which prohibited using them here and rather I relied on checking the structure by introspecting the AST. I will write up a separate issue about that for discussion.

yamaguchi1024:
Correct me if I'm wrong, but my understanding of the motivation of this primitive is that you want to express sth like:
```
for j in seq(0, min(k, n - i - 1)):
  ...
```
The iteration range of `j` depends on the range of `i`, so cutting/splitting the `i` loop will change the result of the "min" eval. However this is not expressible in Exo, so you'll end up having to either write the following instead.
```
for j in seq(0, k):
  if j < n - i - 1:
     ...
```
or
```
for j in seq(0, n - i - 1):
  if j < k:
     ...
```
In either case, after cutting/splitting the `i` loop, you want to simplify the `j` loop to remove the unnecessary iteration.

Smaller comments:
- Why do you restrict `lo` to be 0? There's no inherent reason that it cannot be non-zero, right? Just for implementation simplicity?
- empose -> impose?

SamirDroubi:
Actually, now that I think more about this. If the expression is in the form i < e you should be able to cut the loop at e and run dead code elimination on both loop bodies. This can completely be implemented in the standard library.

PR #566, opened 2024-01-31
Fix bug in inline_assign safety check

SamirDroubi:
* We cannot allow inlining an assign if some buffer in the rhs is read, but also written at some point after the assignment.


yamaguchi1024 (approved):

SamirDroubi:
When I made these checks too strict, it caused some usages to no longer work in BLAS. I will rewrite the safety check for this operation to make it as exact as possible. There is another bug I discovered in its implementation which I will also fix.

yamaguchi1024 (changes requested):
I think the correct thing to do here is not to "prohibit users from doing inline_assign when the expression is written at some point,"  but to "not delete the assign statement when the expression is written." So that it preserves the output behaviour.

yamaguchi1024:
Closing, this and #612 needs to be fixed properly at some point

PR #568, opened 2024-02-02
Forward when updating predicates in simplify

SamirDroubi:


src/exo/LoopIR_scheduling.py
```diff
@@ -3227,9 +3227,12 @@ def __init__(self, proc):
         super().__init__(proc)
 
         # might need to update IR with predicate changes
-        if new_preds := self.map_exprs(self.ir.preds):
-            # TODO KQ: is this line covered? do we not need to forward here?
-            self.ir = self.ir.update(preds=new_preds)
+        new_preds = self.map_exprs(self.ir.preds)
+        if new_preds:
+            self.ir, fwd = (
+                ic.Cursor.create(self.ir)._child_block("preds")._replace(new_preds)
```
skeqiqevian, line 3233: @SamirDroubi Do you often have cursors to predicates? If so, then they wouldn't forward under this implementation. If not, then maybe this is OK.

skeqiqevian (changes requested):
Just wanted to double check that this is the forwarding behavior you intended

src/exo/LoopIR_scheduling.py
SamirDroubi: No, I don't have cursors to predicates ever. The main issue was that other cursors weren't forwarding because of this.

tests/test_cursors.py
```diff
@@ -158,6 +158,19 @@ def foo(n: size, m: size):
     assert str(foo1.forward(stmt)._impl._node) == golden
 
 
+def test_simplify_predicates_forwarding():
+    @proc
+    def foo(n: size):
+        assert n >= 1
```
yamaguchi1024, line 164: If I understand correctly, this PR is fixing the forwarding of simplify when asserts get simplified, right? Is `n >= 1` gonna get simplified even further? My question is whether this test is actually testing the condition `if new_preds: ...`

skeqiqevian: `simplify` modifies all expressions, regardless of whether their actual contents actually change. 

The issue before was that `simplify` resulted in a new `proc` due to changes to the predicates, but the forwarding function didn't update with this information, so provenance was messed up because the forwarding function's final proc was different form the actual final proc.

yamaguchi1024: Ah, got it. Looks good to me

yamaguchi1024 (approved):

skeqiqevian (approved):
LGTM

PR #571, opened 2024-02-02
Fix bug in stage_mem

SamirDroubi:
* When wrapping load/stores with guard, the code was doing a roundabout
    approach to get the block of the load/store which doesn't work when
    the load stage is simply a statement and not a loop nest. The correct
    way to get such a block is to simply use as_block()

### DIFF
--- a/tests/test_schedules.py
+++ b/tests/test_schedules.py
@@ -2603,6 +2603,20 @@ def test_stage_mem_out_of_bound_block(golden):
     assert str(axpy) == golden
 
 
+def test_stage_mem_out_of_bound_point(golden):
+    @proc
+    def foo(n: size, m: size, x: f32[n], y: f32[n]):
+        assert m >= n
+        for i in seq(0, m):
+            if i < n:
+                y[i] = x[i]
+            if i < n:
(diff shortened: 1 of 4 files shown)


yamaguchi1024:
LGTM, but @skeqiqevian should take a look

tests/test_schedules.py
```diff
@@ -2603,6 +2603,20 @@ def axpy(n: size, x: f32[n], y: f32[n]):
...
 
 
+def test_stage_mem_out_of_bound_point(golden):
+    @proc
+    def foo(n: size, m: size, x: f32[n], y: f32[n]):
+        assert m >= n
+        for i in seq(0, m):
+            if i < n:
+                y[i] = x[i]
+            if i < n:
```
skeqiqevian, line 2613: Is there reason we need two `if i < n: y[i] = x[i]` stmts for this test?

SamirDroubi: That's a good point. I can't think of why I wrote it this way lol.

SamirDroubi: Just fixed it.

PR #572, opened 2024-02-03
Change DoEliminateDeadLoop Check

SamirDroubi:
* Change the check from `lo >= hi == True` into `lo >= hi`. The former
    condition wouldn't converge in the z3 solver for the added tests.
* Refactored the Check_ExprBound to directly use strings instead of an
enum. This is generally how we have represented those operations within
the compiler and it is less code to write when calling the check.
* Deleted a duplicated DoEliminateIfDeadBranch function.

### DIFF
--- a/src/exo/LoopIR_scheduling.py
+++ b/src/exo/LoopIR_scheduling.py
@@ -29,7 +29,6 @@ from .new_eff import (
     Check_IsDeadAfter,
     Check_IsIdempotent,
     Check_ExprBound,
-    Check_ExprBound_Options,
     Check_Aliasing,
 )
 from .range_analysis import IndexRangeEnvironment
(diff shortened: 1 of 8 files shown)


yamaguchi1024 (approved):
> The former condition wouldn't converge in the z3 solver for the added tests.

omg really? So cryptic..

src/exo/LoopIR_scheduling.py
```diff
@@ -3808,7 +3778,7 @@ def guard_wrapper(body):
         # the forwarded version. However, in all the current callees, the
         # statement would have been just constructed and if you try to forward
         # you get an error.
-        ir, fwd_wrap = ctxt_stmt_c.parent().body()._wrap(guard_wrapper, "body")
+        ir, fwd_wrap = ctxt_stmt_c.as_block()._wrap(guard_wrapper, "body")
```
yamaguchi1024, line 3781: I think this change is from another PR

SamirDroubi: Yeah, when I merge them inorder. This will go away from this PR

yamaguchi1024: sg

PR #575, opened 2024-02-05
Unify index inequalities with different ops

SamirDroubi:
* Since index inequalities are operating on integers, we can simply rewrite all of them from the form: `lhs op rhs` to `0 < expr`. Then, we can perform unification on the two resulting expressions. This will allow for more unification oppurtunities.
* We have previously implemented a subcase of this, but required the two operations to be the same. This commit generalizes this idea to unify two inequalities of arbitrary ops against each other. Equalities can still only be unified if the other operation is also an equality.

### DIFF
--- a/src/exo/LoopIR_unification.py
+++ b/src/exo/LoopIR_unification.py
@@ -1096,6 +1096,8 @@ class Unification:
         elif e.op == ">=":
             # rewrite `lhs >= rhs` into `0 < lhs - rhs + 1`
             return add1(sub(e.lhs, e.rhs))
+        else:
+            assert False, f"Unreachable op found: {e.op}"
 
     def unify_e(self, pe, be):
         if pe.type.is_indexable() != be.type.is_indexable() or (pe.type == T.bool) != (
--- a/src/exo/LoopIR_unification.py
+++ b/src/exo/LoopIR_unification.py
@@ -1075,6 +1075,28 @@ class Unification:
                 f"type {bt} (@{bnode.srcinfo})"
             )
 
+    @staticmethod
+    def comparision_to_unification_expr(e):
+        def sub(lhs, rhs):
+            return LoopIR.BinOp("-", lhs, rhs, T.index, null_srcinfo())
+
+        def add1(e):
+            one = LoopIR.Const(1, T.int, null_srcinfo())
+            return LoopIR.BinOp("+", e, one, T.index, null_srcinfo())
+
+        if e.op == "<" or e.op == "==":
+            # rewrite `lhs op rhs` into `0 op rhs - lhs`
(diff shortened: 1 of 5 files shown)


src/exo/LoopIR_unification.py
```diff
@@ -1120,19 +1142,23 @@ def unify_e(self, pe, be):
         elif isinstance(pe, LoopIR.USub):
             self.unify_e(pe.arg, be.arg)
         elif isinstance(pe, LoopIR.BinOp):
+            exprs = [pe.rhs, pe.lhs, be.rhs, be.lhs]
+            if pe.op in comparision_ops and all(e.type.is_indexable() for e in exprs):
+                inequality_ops = comparision_ops - {"=="}
+                if pe.op == be.op or (
+                    pe.op in inequality_ops and be.op in inequality_ops
+                ):
```
yamaguchi1024, line 1150: I think this branch is taken when unifying index equality operations like `if a + 1 == b : ...` and `if n == m: ...`, and it seems you're not handling the unification of LHS

src/exo/LoopIR_unification.py
```diff
@@ -1075,6 +1075,28 @@ def unify_types(self, pt, bt, pnode, bnode):
...
+            return sub(e.rhs, e.lhs)
+        elif e.op == "<=":
+            # rewrite `lhs <= rhs` into `0 < rhs - lhs + 1`
+            return add1(sub(e.rhs, e.lhs))
+        elif e.op == ">":
+            # rewrite `lhs > rhs` into `0 < lhs - rhs`
+            return sub(e.lhs, e.rhs)
+        elif e.op == ">=":
+            # rewrite `lhs >= rhs` into `0 < lhs - rhs + 1`
+            return add1(sub(e.lhs, e.rhs))
```
yamaguchi1024, line 1098: You probably want `else` here just for sanity

yamaguchi1024:
This is a great improvement for unification!

src/exo/LoopIR_unification.py
SamirDroubi: Those would go through the first case in `comparision_to_unification_expr`. They will be rewritten into `0 == b - (a + 1)` and `0 == m - n`, then we will unify `b-(a+1)` and `m-n`.

src/exo/LoopIR_unification.py
SamirDroubi: Fixed.

src/exo/LoopIR_unification.py
yamaguchi1024: Ah gocha

yamaguchi1024 (approved):

PR #576, opened 2024-02-06
Cursor forwarding supports many-to-one merging.

skeqiqevian:
This has been implemented in the `mult_loops`, `join_loops` and `fuse_loop` operations. 

I still have to implement it for `fuse_if` and `merge_writes`, but so far the proof of concept seems reasonable. 

Depends on #546


yamaguchi1024:
@skeqiqevian What's the status of this PR? Looks like it'll be nice to merge this

skeqiqevian:
Whether we want to pursue this PR depends on our decision for #582. If we return the relevant cursors we need to, then we can make conservative forwarding decisions and no longer need this PR.

yamaguchi1024:
Isn't it still beneficial to liberally forward cursors when we aren't returning cursors though?

skeqiqevian:
Not if the liberal forwarding decisions are confusing for users to understand though. Considering that relevant cursors will offer the convenience feature that we are currently missing, I think it makes sense to try to make forwarding as unambiguous as possible.

yamaguchi1024:
Closing due to inactivity

PR #579, opened 2024-02-19
Validate the perfectness of divide_loop with z3

### LINKED ISSUE #533: divide_loop should validate perfectness with a real solver
`divide_loop`'s `perfect` argument is an important example of a scheduling directive that performs some custom analyses to check for validity. It currently has ad-hoc checks for constant extents, or for static asserts of the trivial form `assert size%k == 0`. It does not employ a real solver, so it does not generalize even to simple extensions like `assert size1%k == 0 and size2%k == 0`.

Directives like this should be much more powerful in their ability to use static asserts and other conditions in their reasoning.

That could mean:

- Figuring out how `perfect=True` can be made to rely entirely on existing analysis checks
- Adding a simple lowering to an existing solver (likely Z3) for performing these types of queries

The latter is probably useful infrastructure that could be used by many scheduling primitives and higher-level automations, but `divide_loop` seems like a reasonable motivating example with which to start.

SamirDroubi:
Fixes #533.

### DIFF
--- a/src/exo/new_eff.py
+++ b/src/exo/new_eff.py
@@ -2121,7 +2121,7 @@ def Check_ExprBound(proc, stmts, expr, op, value, exception=True):
         query = ADef(e < AInt(value))
         err_msg = f"greater than {value}"
     elif op == "==":
-        query = ADef(e=AInt(value))
+        query = AEq(e, AInt(value))
         err_msg = f"equal to {value}"
     else:
         assert False, "Bad case"
--- a/src/exo/new_eff.py
+++ b/src/exo/new_eff.py
@@ -2121,7 +2121,7 @@ def Check_ExprBound(proc, stmts, expr, op, value, exception=True):
         query = ADef(e < AInt(value))
         err_msg = f"greater than {value}"
     elif op == "==":
-        query = AEq(e, AInt(value))
+        query = ADef(AEq(e, AInt(value)))
         err_msg = f"equal to {value}"
     else:
         assert False, "Bad case"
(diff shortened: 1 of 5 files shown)


src/exo/new_eff.py
```diff
@@ -2121,7 +2121,7 @@ def Check_ExprBound(proc, stmts, expr, op, value, exception=True):
         query = ADef(e < AInt(value))
         err_msg = f"greater than {value}"
     elif op == "==":
-        query = ADef(e=AInt(value))
+        query = AEq(e, AInt(value))
```
skeqiqevian, line 2124: What's the difference between querying `ADef(expr)` and just querying `expr`? I notice all the other cases wrap their queries with a `ADef`, and I'm wondering if `ADef(AEq(e, AInt(value)))` is different from just `AEq(e, AInt(value))`

skeqiqevian (changes requested):
Had one point of confusion, but otherwise seems good!

src/exo/new_eff.py
SamirDroubi: Good catch, I am actually not sure. I will query `ADef(AEq(e, AInt(value)))` just for consistency. Should be fixed now.

skeqiqevian: @yamaguchi1024 do you know?

yamaguchi1024: IIUC, having `ADef` around will ensure that the statement is definitely true (i.e. the reachability of location sets is also checked). If you don't have `ADef`, it would mean "there **may** be some input values where this statement is true".

skeqiqevian (approved):
LGTM

Issue #541, opened 2023-12-19
Proposal for Returning Relevant Cursors


SamirDroubi:
**Problem**:

We have been talking about having scheduling ops support the idea of "Returning relevant cursors". For example, when using `stage_mem`, we generate an allocation (which we might wanna further schedule to for example set the memory), and at least one of a load stage or store stage:
```
def foo(n: size, A: i8[n] @ DRAM):
    assert n % 4 == 0
    for io in seq(0, n / 4):
        for ii in seq(0, 4):
            A[4 * io + ii] += 0.2
```

```
foo = stage_mem(foo, "for ii in _:_", "A[io*4:io*4+4]", "tile")
```

```
def foo(n: size, A: i8[n] @ DRAM):
    assert n % 4 == 0
    for io in seq(0, n / 4):
        tile: i8[4] @ DRAM      <----- Allocation
        for i0 in seq(0, 4):        <----- Load Stage
            tile[i0] = A[i0 + 4 * io]
        for ii in seq(0, 4):
            tile[ii] += 0.2
        for i0 in seq(0, 4):        <------ Store Stage
            A[i0 + 4 * io] = tile[i0]
```
It would be convenient if scheduling operations returned cursors to those relevant pieces of the procedure. Currently the alternative to this is using the argument cursor as an anchor and using navigation to get to such cursors.

**Proposals**:
_Proposal 1_: Return a dataclass algonside the proc. The dataclass contains fields which are the relevant cursors.

Definition goes along with the scheduling operations definition, serves as a documentation of what can be returned:
```
@dataclass
class stage_mem_cursors:
        alloc: AllocCursor = InvalidCursor()
        load_stage: AssignCursor | LoopCursor = InvalidCursor()
        store_stage: AssignCursor | Reduce_Crursor | LoopCursor = InvalidCursor()
```

Usage:
```
foo, cursors = stage_mem(foo, "for ii in _:_", "A[io*4:io*4+4]", "tile")
load_stage = cursors.load_stage # Save for later
foo, cursors = set_memory(foo, cursors.alloc, DRAM_STATIC)
```

Advantages:
1. Explicit API
2. Uniform return values convention `(Procedure, dataclass of relevant cursors)`
3. It is easy to be forward compatiable and add more cursors in the future

Disatvantages:
1. Have to define a dataclass per scheduling op
2. Might be too verbose: `cursors.load_stage`
[...]

PR #582, opened 2024-02-24
Relevant Cursors Feature

SamirDroubi:


SamirDroubi:
Implementing #541

SamirDroubi:
@skeqiqevian I will aim to write the documentation for everything from the top through `mult_dim`.

src/exo/API_scheduling.py
```diff
@@ -862,6 +948,10 @@ def reorder_stmts(proc, block_cursor):
 
     args:
         block_cursor    - a cursor to a two statement block to reorder
+        rc              - bool (whether to return relevant cursors)
+
+    relevant cursors:
+        block           - a block containing the two statements after the reordering.
```
SamirDroubi, line 954: This is not specific to `reorder_stmts`, but rather for the broader API design: 

I would prefer the names of the fields in the returned data-structure to be as minimal as possible to make the user-code as simple as possible. In this example, I would rather name it `block` rather than `block_cursor`. The `_cursor` suffix generally wouldn't give me any new information. However, this is conflicts with the API naming convention for parameters; parameters are often called `{name}_cursror`. I would argue that the parameters should be renamed to just `{name}` since they generally accept a broader set of arguments than just cursors: patterns. You could argue that the pattern itself is somewhat a cursor, but I am not sure if it is clearer to suffix the parameters with `_cursor`.

skeqiqevian: I agree that the return names don't need "_cursor" because we are only returning cursors. Currently undecided on the scheduling operation arguments, but since those were never consistent to begin with, I don't think we need to worry about conflicting with it?

src/exo/API_scheduling.py
```diff
@@ -2199,6 +2537,15 @@ def specialize(proc, block_cursor, conds):
         block_cursor    - cursor pointing to the block to duplicate/specialize
         conds           - list of strings or string to be parsed into
                           guard conditions for the
+        rc              - bool (whether to return relevant cursors)
+
+    relevant_cursors:
+        if              - a cursor to the outermost if-statement
+        TODO: should we return each of the subcases? There's a natural navigation
```
skeqiqevian, line 2544: Some scheduling operations seem like they should maybe be split into multiple scheduling operations. There's `specialize`, `lift_scope` and `eliminate_dead_code`

src/exo/API_scheduling.py
```diff
@@ -1512,6 +1726,12 @@ def inline_window(proc, winstmt_cursor):
 
     args:
         winstmt_cursor  - cursor pointing to the WindowStmt to inline
+        rc              - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        exprs           - cursors to all replaced window expressions
+        TODO: unsure of this one, since it's hard to distinguish
```
skeqiqevian, line 1733: This is one of the cases where it might be worth returning modified cursors because after the edit, there's no way to identify what was previously using the variable from the window statement.

src/exo/API_scheduling.py
```diff
@@ -766,6 +792,36 @@ def parse_arg(a):
...
+# Another option is to use a `NamedTuple`, but one drawback of this is that we
+# cannot make it a subtype of some base class like we did below.
+
+
+@dataclass
+class CursorsSet:
+    pass
+
+
+def make_cursors_set(sched_op, fields):
```
gilbo, line 810: This only appears to be invoked for `simplify()` right now.  Is the implementation only half complete?

SamirDroubi: We didn't want to actually implement the feature until we converge on what the implementation should achieve. Just to avoid redoing work. This was just supposed to be a sample of using this small infrastructure for supporting the feature.

src/exo/API_scheduling.py
gilbo: I'm supportive of this change.  Although we should have a better scheme for "type" annotation / argument processors.

Let's defer changing those names to a later PR focused on revamping the argument processors.  As is, the longer parameter names rarely appear in code.

src/exo/API_scheduling.py
```diff
@@ -890,6 +997,10 @@ def commute_expr(proc, expr_cursors):
 
     args:
         expr_cursors - a list of cursors to the binary operation
+        rc           - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        exprs        - a tuple of cursors to the commuted expressions
```
gilbo, line 1003: This seems fine, but dropping comment because I hadn't seen this scheduling op before.

Should it be called `commute_exprs` if it takes a list of cursors?  Also, should we just stick to the convention that it should take a single cursor, but could be applied repeatedly?

gilbo: Same comment for similar expr ops added after I last went through this file

src/exo/API_scheduling.py
```diff
@@ -1253,6 +1423,10 @@ def expand_dim(proc, buf_cursor, alloc_dim, indexing_expr):
                           of the new buffer dimension.
         indexing_expr   - (string) an expression to index the newly
                           created dimension with.
+        rc              - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        alloc           - a cursor to the modifed alloc
```
gilbo, line 1430: Not pertinent to this PR so much, but there is a TODO to rename this op.  I agree with `add_dim` except furthermore suggest `add_dims` since the plural makes it more clear that you're talking about the dimensions in aggregate rather than modifying an individual dimension.

src/exo/API_scheduling.py
```diff
@@ -1277,6 +1451,10 @@ def rearrange_dim(proc, buf_cursor, permute_vector):
         buf_cursor      - cursor pointing to an Alloc statement
                           for an N-dimensional array
         permute_vector  - a permutation of the integers (0,1,...,N-1)
+        rc              - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        alloc           - a cursor to the modifed alloc
```
gilbo, line 1458: Similar to the above comment, this op should be renamed to `rearrange_dims` most likely

src/exo/API_scheduling.py
```diff
@@ -1340,6 +1522,10 @@ def mult_dim(proc, alloc_cursor, hi_dim_idx, lo_dim_idx):
         alloc_cursor    - cursor to the allocation to divide a dimension of
         hi_dim_idx      - the index of the higher order dimension to multiply
         lo_dim_idx      - the index of the lower order dimension to multiply
+        rc              - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        alloc           - a cursor to the modifed alloc
```
gilbo, line 1528: sim. to the above, should this be `mult_dims`?

src/exo/API_scheduling.py
```diff
@@ -1415,6 +1609,10 @@ def sink_alloc(proc, alloc_cursor):
```
gilbo, line 1609: Typo above "statemenet"

gilbo: Probably blame me for writing it, lol

src/exo/API_scheduling.py
gilbo: I support this

gilbo: Actually, this may be tricky...

1) The alternative would be to use an introspection feature to get all the window use-sites before doing this operation and then count on forwarding to preserve the identities of all those use-sites.

2) The use-sites are not a homogeneous type of cursor, since the window could be used as both an lvalue or an rvalue.  We need to think more carefully about what exactly would get returned here, and how it relates to inspection features...

src/exo/API_scheduling.py
```diff
@@ -1648,6 +1887,14 @@ def divide_loop(proc, loop_cursor, div_const, new_iters, tail="guard", perfect=F
...
                           be zero (i.e. there is no tail).  You will get an
                           error if the compiler cannot verify this fact itself.
+        rc              - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        outer_loop      - a cursor to the new outer loop
+        inner_loop      - a cursor to the new inner loop
+        ?main_guard     - a cursor to the guard inside the main loop (only for "guard")
+        ?tail_guard     - a cursor to the guard enclosing the tail loop (only for "cut_and_guard")
+        ?tail           - a cursor to the tail loop (only for "cut" and "cut_and_guard")
```
gilbo, line 1897: Can tail and tail-guard just be collapsed?  Likewise, do we need main-guard, or can one just use the body of the inner-loop?

src/exo/API_scheduling.py
```diff
@@ -1707,6 +1958,10 @@ def join_loops(proc, loop1_cursor, loop2_cursor):
     args:
         loop1_cursor     - cursor pointing to the first loop
         loop2_cursor     - cursor pointing to the second loop
+        rc               - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        loop             - a cursor to the new joined loop
 
     rewrite:
         `for i in seq(lo, mid):`
```
gilbo, line 1967: If every loop is 0-based on the bottom end, then this example code is bogus.  This is probably my own fault, but we should try to fix it to be accurate

src/exo/API_scheduling.py
```diff
@@ -1788,6 +2052,11 @@ def reorder_loops(proc, nested_loops):
                           variable of the inner loop.  An optional '#int'
                           can be added to the end of this shorthand to
                           specify which match you want,
+        rc              - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        outer_loop      - a cursor to the new outer loop
+        inner_loop      - a cursor to the new inner loop
```
gilbo, line 2059: Should these (and elsewhere) be `outer` and `inner` in keeping with your stated preference for shorter names?

src/exo/API_scheduling.py
```diff
@@ -1887,6 +2164,13 @@ def inline_assign(proc, alloc_cursor):
     """
     Inlines [alloc_cursor] into any statements where it is used after this assignment.
```
gilbo, line 2165: This is mis-written.  The cursor is an assignment cursor, not an allocation

src/exo/API_scheduling.py
```diff
@@ -1911,6 +2195,12 @@ def lift_reduce_constant(proc, block_cursor):
```
gilbo, line 2195: The name of this operation should use the word "factor" as in "factor out"

src/exo/API_scheduling.py
```diff
@@ -2056,12 +2362,17 @@ def fuse(proc, stmt1, stmt2, unsafe_disable_check=False):
...
+    when the block of statements in question can be proven to be idempotent. If the
+    loop is not guaranteed to execute at least once, we wrap a guard around its body.
 
     args:
         loop_cursor     - cursor pointing to the loop to remove
+        rc              - (bool) whether to return relevant cursors
+
+    relevant cursors:
+        block           - a cursor to the block of statements that was inside the loop
+        ?guard          - a cursor to the guard around the loop body
```
gilbo, line 2375: Can we always return a `guard` cursor?  (but change the name) and it just happens that the two cursors here coincide when there is no guard?

src/exo/API_scheduling.py
```diff
@@ -2091,6 +2403,12 @@ def add_loop(
                           wrap the block in a `if iter_name == 0: block`
                           condition; in which case idempotency need not
                           be proven.
+        rc              - (bool) whether to return relevant cursors
+
+    relevant_cursors:
+        loop            - a cursor to the new loop
+        block           - a cursor to the block of statements that was wrapped inside the loop
+        ?guard          - a cursor to the guard around the loop body
```
gilbo, line 2411: Similar comment to the above

src/exo/API_scheduling.py
```diff
@@ -2141,6 +2463,16 @@ def lift_scope(proc, scope_cursor):
...
+        rc                 - bool (whether to return relevant cursors)
+
+    relevant cursors:
+        ?if                - a cursor to the lifted if statement
+        ?if_block          - a cursor to the block within the if-clause
+        ?else_block        - a cursor to the block within the else-clause
+        ?for               - a cursor to the lifted for statement
+        TODO: In trying to decide relevant cursors for this scheduling operation,
+        I think that this scheduling operation should really be split into two
+        operations: lift_if and lift_for.
```
gilbo, line 2475: It used to be that way; I agree

src/exo/API_scheduling.py
```diff
@@ -2170,6 +2502,12 @@ def eliminate_dead_code(proc, stmt_cursor):
 
     args:
         stmt_cursor       - cursor to the if or for statement
+        rc                - bool (whether to return relevant cursors)
+
+    relevant cursors:
+        ?block             - block within the if statement which is always executed
+        TODO: In trying to decide relevant cursors for this scheduling operation,
+        I think that this scheduling operation should really be split into two.
```
gilbo, line 2510: I think you can omit any relevant cursors on the principle that this is very similar to a "delete statement"

src/exo/API_scheduling.py
gilbo: I don't understand why `specialize` should be split into something more primitive.  I think either of the options of "just return the one cursor" or "return the one outer cursor and also return a tuple of cursors, one for the body of each branch" are both reasonable specs for `rc` here.

yamaguchi1024:
I wasn't following this discussion and PR closely, but it would be nice to come back to this and discuss some plans for merging it at some point.

yamaguchi1024:
Closing due to inactivity. It's unfortunate that we didn't get to merge this PR, but I personally think that the primitive interface is simpler without returning relevant cursors 🤷‍♀️

PR #584, opened 2024-03-05
Implement delete_pass using cursors

SamirDroubi:


yamaguchi1024:
I delegate this to @skeqiqevian

src/exo/LoopIR_scheduling.py
```diff
@@ -2716,24 +2716,34 @@ def DoDeleteConfig(proc_cursor, config_cursor):
...
+    def result(self, **kwargs):
+        return api.Procedure(
+            self.ir, _provenance_eq_Procedure=self.provenance, _forward=self.fwd
+        )
+
+    def delete_cursor(self, sc):
+        self.ir, fwd = self.fwd(sc)._delete()
+        self.fwd = _compose(fwd, self.fwd)
 
     def map_s(self, sc):
```
skeqiqevian, line 2733: The whole point of cursors is that this implementation should be a few lines. You can pattern match to get cursors to all `pass` statements, and then you can just call delete on them (and delete their parent if they are the only statement in the loop). You don't need to implement any of the `Cursor_Rewrite` class's traversal functions.

skeqiqevian (changes requested):
I would restructure this implementation to be easier to read, see my comment above.

src/exo/LoopIR_scheduling.py
SamirDroubi: Done.

skeqiqevian: I feel like the code could still be condensed, but maybe I'm wrong. I'll try to do so in a bit.

skeqiqevian (approved):
Assuming tests pass, LGTM

PR #585, opened 2024-03-06
Rerwite extract_subproc

SamirDroubi:
* Rewrite extract_subproc using cursors.
* Allow extracting blocks.
* Order parameters by the order in which they appear in the proc.
* Prepare order parameter to be deprecated.

### DIFF
--- a/src/exo/API_scheduling.py
+++ b/src/exo/API_scheduling.py
@@ -1003,13 +1003,15 @@ def bind_expr(proc, expr_cursors, new_name):
 
 
 @sched_op([NameA, StmtCursorA, DictA])
-def extract_subproc(proc, subproc_name, body_stmt, order=dict()):
+def extract_subproc(proc, subproc_name, stmt, order=dict()):
     """
     Documentation
     """
-    stmt = body_stmt._impl
-    passobj = scheduling.DoExtractMethod(proc, subproc_name, stmt, order)
-    return passobj.result(), passobj.subproc()
+
+    ir, fwd, subproc_ir = scheduling.DoExtractMethod(stmt._impl, subproc_name, order)
+    proc = Procedure(ir, _provenance_eq_Procedure=proc, _forward=fwd)
+    subproc = Procedure(subproc_ir)
+    return proc, subproc
 
 
 @sched_op([CallCursorA])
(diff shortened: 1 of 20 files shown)


SamirDroubi:
Code diff related to delete_pass is from #584.

src/exo/API_scheduling.py
```diff
@@ -1001,14 +1002,46 @@ def bind_expr(proc, expr_cursors, new_name):
...
-    Documentation
+@sched_op([NameA, BlockCursorA, DictA])
+def extract_subproc(proc, subproc_name, block, order=dict()):
     """
-    stmt = body_stmt._impl
-    passobj = scheduling.DoExtractMethod(proc, subproc_name, stmt, order)
-    return passobj.result(), passobj.subproc()
+    Extract a block as a subprocedure with the name `subproc_name`.
+
+    # TODO: deprecate `order` and swap the order of `subproc_name` and `block`.
```
SamirDroubi, line 1010: Once we confirm the CI passes and we review this code, I plan to do this which will break the BLAS code.

yamaguchi1024: Why do you want to deprecate `order`? There will be no way of permuting the argument ordering otherwise.

yamaguchi1024: I'm mildly against deprecating order

SamirDroubi: It is not a valuable parameter if the implementation is written properly. 

Previously, it was being used as a workaround because the parameters of the subproc would be created in a different order than the parent proc. Now, that they follow the same order. There is no good reason for it. If you notice, currently, order is not being used in the implementation and all the tests pass.

src/exo/API_scheduling.py
```diff
@@ -1002,14 +1002,46 @@ def bind_expr(proc, expr_cursors, new_name):
 # Sub-procedure Operations
 
 
-@sched_op([NameA, StmtCursorA, DictA])
-def extract_subproc(proc, subproc_name, body_stmt, order=dict()):
+@sched_op([NameA, BlockCursorA, DictA])
+def extract_subproc(proc, subproc_name, block, order=dict()):
```
yamaguchi1024, line 1006: Add an optional boolean argument for putting assertions

yamaguchi1024 (approved):

PR #586, opened 2024-03-06
Deprecate bound_and_guard operation

### LINKED ISSUE #570: Deprecate bound_and_guard usage in sgemm.py
Port the right panel implementation from BLAS GEMM to apps/x86/sgemm.py. bound_and_guard doesn't have a safety check, and I just want to ensure that our code doesn't depend on unsafe rewrite before making analysis changes.

This is not a blocker, but assigning it to @SamirDroubi so that we don't forget

SamirDroubi:
Fixed #570

### DIFF
--- a/src/exo/API_scheduling.py
+++ b/src/exo/API_scheduling.py
@@ -1002,14 +1002,47 @@ def bind_expr(proc, expr_cursors, new_name):
 # Sub-procedure Operations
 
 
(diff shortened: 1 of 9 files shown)


yamaguchi1024:
Can you deprecate DoStageWindow as well in this PR?

yamaguchi1024 (approved):

PR #588, opened 2024-03-09
Fix various bugs in memory and type setting

SamirDroubi:
* Ensure that memory is never None for allocation with numerical data types.
* Refactor environment extraction and deprecate that logic from DoStageMem.
* Ensure that all internal types get lowered to API types.


yamaguchi1024 (changes requested):
The `mem` attribute will be mandatory with this PR, so it should also make `mem` non-optional in the LoopIR
```
         | Alloc( sym name, type type, mem? mem )
         | Free( sym name, type type, mem? mem )
```

SamirDroubi:
> The `mem` attribute will be mandatory with this PR, so it should also make `mem` non-optional (remove `?`) in the LoopIR (Alloc, Free, and fnarg).

I think you are right about alloc/free, but with fnarg you can have size/index parameters, and we haven't defined what the memory for those are before. We can potentially give them a `Memory` annotation or a `DRAM`. We can potentially look into what the C standard says about those parameters and mimic it.

yamaguchi1024:
@SamirDroubi I see, I think we should keep `fnarg`'s `mem` optional then

yamaguchi1024 (approved):
Great to see this stuff getting cleaned up

PR #590, opened 2024-03-10
extract_subproc assigns memories to subproc params

SamirDroubi:


src/exo/LoopIR_scheduling.py
```diff
@@ -2722,10 +2722,10 @@ def make_closure():
         # Construct the parameters and arguments
         args = []
         fnargs = []
-        for sym, typ, _ in sym_env:
+        for sym, typ, mem in sym_env:
             if sym in body_symbols:
                 args.append(LoopIR.Read(sym, [], typ, info))
-                fnargs.append(LoopIR.fnarg(sym, typ, None, info))
+                fnargs.append(LoopIR.fnarg(sym, typ, mem, info))
```
yamaguchi1024, line 2728: Isn't this implicitly assigning `mem` to control types, too? I think having `mem = DRAM` on `size` and `index` don't really make sense (even though they are implicit in the LoopIR and are never printed).

SamirDroubi: It will use whatever memory type was assigned to these `size`/`index` variables which should be `None`.

SamirDroubi:
Let's hold-off on merging this. I found an unrelated bug when doing code generation for programs where we have used extract_proc (it seems due to incorrect symbols management). I will just fix it within this same PR.

src/exo/LoopIR_scheduling.py
yamaguchi1024: Can you add an assert to check that it's actually None?

SamirDroubi: Actually, currently the behavior is to set all the argument to DRAM if they don't have a memory specified from the front-end code: https://github.com/exo-lang/exo/blob/9d8e69cdaafdc45646ab592f0ed77ca9aa4719fc/src/exo/typecheck.py#L43

SamirDroubi: It also ultimately gets suppressed at codegen: https://github.com/exo-lang/exo/blob/9d8e69cdaafdc45646ab592f0ed77ca9aa4719fc/src/exo/LoopIR_compiler.py#L538

I think it is safe to keep it as DRAM and it also means that we can relax the `mem?` in the LoopIR definition of `fnarg` to simply `mem`

yamaguchi1024 (approved):

PR #598, opened 2024-03-21
Reorganizing Halide apps and fixed blur's test/documentation

skeqiqevian:
I accidentally reused the same branch so it looks like there's a lot of commits, but really it's just one small commit + merging main.


apps/halide/blur/main.cpp
```diff
@@ -96,40 +96,23 @@ bool write_png_file(
...
 
 int exec_parrot(blurtype func, std::string output_name, int width, int height,
-    uint8_t *parrot) {
-  uint8_t *parrot_blurred_0, *parrot_blurred_1;
-  parrot_blurred_0 =
-      (uint8_t *)malloc(sizeof(uint8_t) * (width + 4) * (height + 4));
-  parrot_blurred_1 =
-      (uint8_t *)malloc(sizeof(uint8_t) * (width + 4) * (height + 4));
+    uint16_t *parrot) {
+  uint16_t *parrot_write = (uint16_t *)malloc(sizeof(uint16_t) * width * height);
```
SamirDroubi: You have a memory leak here. You should add a corresponding free or wrap it with a smart pointer. Another option is to use a std::vector and provide the size of the buffer to the constructor. Then, you can extract the raw pointer with std::vector::data()

apps/halide/blur/main.cpp
```diff
@@ -139,23 +122,22 @@ int exec_parrot(blurtype func, std::string output_name, int width, int height,
...
   int width, height;
 
   if (read_png_file(read_file, buffer, width, height)) {
     printf("width: %d\n", (int)width);
     printf("height: %d\n", (int)height);
 
-    uint8_t *parrot;
-    parrot = (uint8_t *)malloc(sizeof(uint8_t) * (width + 4) * (height + 4));
+    uint16_t *parrot;
+    parrot = (uint16_t *)malloc(sizeof(uint16_t) * (width + 2) * (height + 2));
```
SamirDroubi, line 135: Same memory lake above.

yamaguchi1024 (approved):

apps/halide/blur/main.cpp
```diff
@@ -96,40 +96,23 @@ bool write_png_file(
...
-  parrot_blurred_0 =
-      (uint8_t *)malloc(sizeof(uint8_t) * (width + 4) * (height + 4));
-  parrot_blurred_1 =
-      (uint8_t *)malloc(sizeof(uint8_t) * (width + 4) * (height + 4));
+    uint16_t *parrot) {
+  uint16_t *parrot_write = (uint16_t *)malloc(sizeof(uint16_t) * width * height);
 
   auto start = std::chrono::steady_clock::now();
   int iterations = 100;
   for (int i = 0; i < iterations; i += 4) {
```
SamirDroubi, line 108: Why are you skipping 4 iterations at a time? It seems like this will mean you are doing an incorrect average later on because you are assuming you are running the program `iterations` times.

apps/halide/blur/main.cpp
```diff
@@ -139,23 +122,22 @@ int exec_parrot(blurtype func, std::string output_name, int width, int height,
...
-    uint8_t *parrot;
-    parrot = (uint8_t *)malloc(sizeof(uint8_t) * (width + 4) * (height + 4));
+    uint16_t *parrot;
+    parrot = (uint16_t *)malloc(sizeof(uint16_t) * (width + 2) * (height + 2));
     for (int i = 0; i < height; i++) {
-      memcpy(&parrot[(i + 2) * (width + 4) + 2], &buffer[i * width],
-          sizeof(uint8_t) * width);
+      memcpy(&parrot[(i + 1) * (width + 2) + 1], &buffer[i * width],
+          sizeof(uint16_t) * width);
     }
```
SamirDroubi, line 140: Can we make exec_parrot return the output buffer, then we can add a sanity check for correctness at the end of the file instead of having to run `diff` on the outputs? You can use `memcmp` or directly compare vectors if you use std::vector.

apps/halide/blur/main.cpp
```diff
@@ -96,40 +96,23 @@ bool write_png_file(
 }
 
 typedef void (*blurtype)(
-    void *ctxt, int_fast32_t n, int_fast32_t m, uint8_t *g, const uint8_t *inp);
+    void *ctxt, int_fast32_t n, int_fast32_t m, uint16_t *g, const uint16_t *inp);
 
 int exec_parrot(blurtype func, std::string output_name, int width, int height,
```
SamirDroubi, line 101: Should this function be renamed to `exec_func` or `exec_blur`?

Should all the buffers be renamed from `parrot*` to `inp*/img*`

SamirDroubi (changes requested):

apps/halide/blur/main.cpp
skeqiqevian: oops accident from Yuka's code

SamirDroubi (approved):

PR #599, opened 2024-03-22
Halide unsharp masking

skeqiqevian:
This current schedule gets 15.1s on a 3200x3200 image (the manual Halide schedule gets 13.3s). I will implement the circular buffer optimization separately in a later PR, which will get the Exo schedule to 12.6s.


apps/x86/halide/unsharp/unsharp/unsharp.h
```diff
@@ -0,0 +1,68 @@
+
```
SamirDroubi, line 1: Is there a reason why the codegen results is committed in both: `apps/x86/halide/unsharp/unsharp` and `tests/golden/test_apps/test_unsharp.txt`? I imagine you only need the latter.

skeqiqevian: Oh, that's true. Removed it (and also the blur.c and blur.h files I previously committed).

apps/x86/halide/unsharp/unsharp.py
```diff
@@ -0,0 +1,177 @@
...
+    mm256_prefix_sub_ps,
+    mm256_prefix_div_ps,
+    mm256_prefix_broadcast_ss,
+    # TODO: need constant broadcast and sub/div masked instructions
+]
+
+
+def halide_vectorize(p, buffer: str, loop: str, width: int):
+    loop = get_enclosing_loop_by_name(p, p.find(f"{buffer} = _"), loop)
+    p = vectorize(p, loop, width, "f32", AVX2, avx_f32_insts, rules=[])
```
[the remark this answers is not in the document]
skeqiqevian: True, but I don't think it's that necessary, especially since vectorization currently doesn't affect performance at all. This is mainly a demo that I can mostly match the Halide schedule, so I think it's OK to not use it.

apps/x86/halide/unsharp/unsharp.py
```diff
@@ -0,0 +1,153 @@
...
+
+    return p
+
+
+# from math import exp, pi, sqrt
+# sigma = 1.5
+# kernel = [exp(-x * x / (2 * sigma * sigma)) / (sqrt(2 * pi) * sigma) for x in range(4)]
```
SamirDroubi, line 40: What if you defined this as a:
k0, k1, k2, k3 = kernel

Then, below assign k0, k1, k2, k3 to kernel[0] kernel[1] kernel[2] kernel[3].

src/exo/stdlib/halide_scheduling_ops.py
```diff
@@ -8,32 +8,63 @@
...
+# TODO: add returning relevant cursors to all the scheduling ops
+
+
 def compute_at(proc: Procedure, producer_assign: AssignCursor, target_loop: ForCursor):
     """
     Computes the necessary values of producer at the level of [target_loop]
     in the consumer loop nest by fusing the two loop nests.
 
+    NOTE: This implementation assumes that both loop nests are at the top-level, but
+    it is easy to extend if we replace `get_toplevel_stmt` with an LCA function of
```
SamirDroubi, line 62: There is a get_lca function in inspection.py if that helps.

SamirDroubi (commented):
I took a look at some stuff. I will leave the final approval to @yamaguchi1024 for the Halide ops.

src/exo/stdlib/halide_scheduling_ops.py
skeqiqevian: I think this is fine, mostly because Halide implicitly deals with top-level loop nests anyways with Funcs.

yamaguchi1024 (approved):
lgtm!

skeqiqevian:
@SamirDroubi you're blocking the merge lol

SamirDroubi (approved):

PR #600, opened 2024-03-25
rearrange_dim and delete_pass bug fixes

SamirDroubi:
- delete_pass is a no-op when there are no passes.
    * We have previously used the external pattern matching API to implement
      delete_pass which throws an error when matches aren't found.
    * Now, we use the internal pattern matching which simply returns an
      empty list.
- Fix bug in rearrange_dim value checking.
    * The original check didn't account for the case where the list
      contains all the numbers between 0 and N-1, but is longer than N.


yamaguchi1024:
@skeqiqevian should review this

skeqiqevian (approved):
LGTM

PR #603, opened 2024-03-26
Prevent divide_with_recompute from generating useless inner loops

skeqiqevian:


src/exo/LoopIR_scheduling.py
```diff
@@ -636,24 +636,38 @@ def szop(op, lhs, rhs):
...
-            LoopIR.Const(0, T.index, srcinfo),
-            hi_i,
-            body,
-            LoopIR.Seq(),
-            None,
-            srcinfo,
+    wrapped_inner_loop = True
+    try:
+        Check_ExprEqvInContext(
+            proc, hi_i, [loop], LoopIR.Const(1.0, T.index, loop.srcinfo), [loop]
```
yamaguchi1024, line 642: Wait I'm a bit confused, why is this hardcoding 1.0? Is this a workaround for the specific pattern that you're looking at?

skeqiqevian: I'm specifically trying to avoid generating loops of size 1.

yamaguchi1024: Why is that a problem?

SamirDroubi: I think you want scheduling operations to always generate the same structure. Especially, that a lot of our implementations rely on that. If you are working with a case where you need this, you can probably call `rewrite_expr(proc, outer_loop.hi(), "1")`, then `unroll_loop(proc, outer_loop)`.

yamaguchi1024: Yeah I second that

skeqiqevian: @SamirDroubi I don't love that approach because `unroll_loop` will break all forwarding to code within it. In particular, a lot of Halide code is loop nest around an assign statement, and I would like to maintain forwarding for the assign statement.

SamirDroubi: Ugh you are right. Could we make `unroll_loop` forward in that edge case since it might be a desirable behavior in other cases as well? I actually have a bunch of cases where I unroll loops of size 1 as well.

yamaguchi1024: Will it be crazy to forward the loop body to the first unrolled loop body?

yamaguchi1024: Yeah, there's no perfect solution. I think we should minimize surprise in general.

yamaguchi1024: divide_with_recompute generating different loop structure for hi == 1 is very surprising. We should weigh that with how surprising the unroll forwarding behavior is.

skeqiqevian: Oh, I guess one solution is rather than modifying the schedulign operation, I could build a lightweight wrapper around `divide_with_recompute` which will only perform the operation if it generates a non-trivial loop by introspecting the inner loop's bound after applying the operation.

SamirDroubi: Instead of using `unroll_loop`, you could also use `remove_loop` whose forwarding behavior is what you want, but you might need to slightly change its implementation to handle this corner case (hi - lo = 1). In this case, you don't need to verify idempotency and you can immediately do the transform.

Although, now `remove_loop` would be correct for two different reasons rather just one.

yamaguchi1024: Changing `remove_loop` will be super hacky too. Generally I think we shouldn't hack in corner cases to scheduling ops implementation

skeqiqevian:
Rather than changing the original scheduling operation, I've decided to just implement a lightweight wrapper around `divide_with_recompute` which will only perform the operation if it generates a non-trivial loop by introspecting the inner loop's bound after applying the operation.

Something like the following:
```
def _divide_with_recompute(proc, loop_cursor: ForCursor, outer_hi, outer_stride: int, new_iters: list[str]):
    """
    Wrapper around `divide_with_recompute` which will not perform the operation
    if it generates a trivial inner loop of size 1.
    """
    temp_proc = divide_with_recompute(proc, loop_cursor, outer_hi, outer_stride, new_iters)
    temp_proc = simplify(temp_proc)
    if is_literal(temp_proc, temp_proc.forward(loop_cursor).body()[0].hi(), 1):
        return proc
    return temp_proc
```

skeqiqevian:
Yes, for now it's only going to be in `halide_scheduling_ops.py`

src/exo/LoopIR_scheduling.py
skeqiqevian: Yeah I considered modifying `remove_loop`, but I think I agree with Yuka about keeping primitives as general/simple as possible.

PR #605, opened 2024-03-27
Scheduling operation for the circular buffer optimization

skeqiqevian:


src/exo/LoopIR_scheduling.py
```diff
@@ -3584,6 +3584,193 @@ def mk_write(c):
...
+        return self.exit_scope()
+
+    def do_s(self, s):
+        bounds = None
+        if isinstance(s, LoopIR.For):
+            bounds = self.do_stmts(s.body)
+            env = {
+                s.iter: (
+                    s.lo.val if isinstance(s.lo, LoopIR.Const) else None,
+                    s.hi.val - 1 if isinstance(s.hi, LoopIR.Const) else None,
```
yamaguchi1024, line 3662: So this doesn't handle loops like
```
for i in seq(m, m + 3):
  ...
```
?

src/exo/LoopIR_scheduling.py
```diff
@@ -3584,6 +3584,193 @@ def mk_write(c):
...
+    def enter_scope(self):
+        self.env.enter_scope()
+        self.bounds.append(None)
+
+    def exit_scope(self) -> Optional[IndexRange]:
+        self.env.exit_scope()
+        return self.bounds.pop()
+
+    def validate_access(self, idx: Optional[int]) -> bool:
+        if self.bounds is None:
```
yamaguchi1024, line 3635: Is this None or []?

src/exo/LoopIR_scheduling.py
```diff
@@ -3584,6 +3584,193 @@ def mk_write(c):
...
+    def do_e(self, e):
+        if isinstance(e, LoopIR.Read) and e.name == self.name:
+            index_rng = index_range_analysis_wrapper(e.idx[self.dim], self.env)
+            self.update_bounds(index_rng)
+        elif isinstance(e, LoopIR.WindowExpr) and e.name == self.name:
+            w_access = e.idx[self.dim]
+            if isinstance(w_access, LoopIR.Interval):
+                lo_rng = index_range_analysis_wrapper(w_access.lo, self.env)
+                hi_rng = index_range_analysis_wrapper(w_access.hi, self.env)
+                self.update_bounds(lo_rng | hi_rng)
```
yamaguchi1024, line 3703: Why is this not using `merge_index_ranges`?

src/exo/LoopIR_scheduling.py
```diff
@@ -3584,6 +3584,193 @@ def mk_write(c):
...
+                self.validate_access(new_bounds.lo)
+            self.bounds[-1] |= new_bounds
+
+    def do_stmts(self, stmts):
+        self.enter_scope()
+        super().do_stmts(stmts)
+        return self.exit_scope()
+
+    def do_s(self, s):
+        bounds = None
```
yamaguchi1024, line 3656: Can you not use the same name for self.bounds and bounds?

src/exo/LoopIR_scheduling.py
skeqiqevian: Fixed to support this. However, currently we cannot handle general for loops with non-constant window loop bounds.

src/exo/LoopIR_scheduling.py
skeqiqevian: It doesn't have to because `index_range_analysis` will always return a valid IndexRange. `merge_index_ranges` is necessary if we don't know that the inputs might be `None` type.

src/exo/LoopIR_scheduling.py
```diff
@@ -3584,6 +3586,193 @@ def mk_write(c):
...
+                    )
+            self.access_window_per_scope[-1] |= new_bounds
+
+    def do_stmts(self, stmts):
+        self.enter_scope()
+        super().do_stmts(stmts)
+        return self.exit_scope()
+
+    def do_s(self, s):
+        print(s, self.access_window_per_scope[-1])
```
yamaguchi1024, line 3645: delete print?

tests/test_schedules.py
```diff
@@ -1132,6 +1132,188 @@ def foo(a: f32 @ DRAM, b: f32 @ DRAM):
...
+#             x[2 * i] = 1.0
+#             x[2 * i + 1] = 2.0
+#         x[N / 2] = 3.0
+
+#     with pytest.raises(SchedulingError, match="Buffer folding not possible"):
+#         foo = fold_buffer(foo, foo.find("x: _"), 0, 2)
+
+#     foo = fold_buffer(foo, foo.find("x: _"), 0, 3)
+#     foo = simplify(foo)
+#     assert str(foo) == golden
```
yamaguchi1024, line 1170:
```
x : R[N]
for i in seq(0, N):
  x[i] = 0.0
x[0] = 1.0
```

tests/test_schedules.py
```diff
@@ -1132,6 +1132,188 @@ def foo(a: f32 @ DRAM, b: f32 @ DRAM):
...
+def test_fold_buffer_sequential_stmts(golden):
+    @proc
+    def foo():
+        x: i8[10]
+        x[0] = 0.0
+        x[2] = 0.0
+        x[3] = 0.0
+        x[2] = 0.0
+        x[7] = 0.0
+        x[5] = 0.0
```
yamaguchi1024, line 1182: This should succeed, right?
```
    @proc
    def foo():
        x: i8[20]
        x[0] = 0.0
        x[2] = 0.0
        x[3] = 0.0
        x[2] = 0.0
        x[7] = 0.0
        x[5] = 0.0
```
`foo = fold_buffer(foo, foo.find("x: _"), 0, 10)`

skeqiqevian:
@yamaguchi1024 Can you approave? 

This change will fail ExoBLAS tests because I fixed a bug with range analysis which previously enabled omitting `exo_floor_div` in some asserts. While those omissions were mathematically correct, it's not something that our current range analysis should be able to handle, so it was only able to make the omission via the bug that I fixed.

yamaguchi1024 (approved):

yamaguchi1024:
Sounds good! @SamirDroubi might want to review, too?

PR #606, opened 2024-03-29
Guard windows definitions with "include guards"

SamirDroubi:
* We need to define exo window structs within the headers we generate
    because those definition depend on the kernel.
* However, this means that including multiple headers generated by Exo
   can cause a compilation error because of duplicate definitions.
* This commit wraps those definitions with "include guards" to avoid
   such conflicts.


SamirDroubi:
This will break the BLAS codegen tests (the last step in the workflow), but we sure verify that it passes the build and correctness tests.

SamirDroubi:
> This will break the BLAS codegen tests (the last step in the workflow), but we sure verify that it passes the build and correctness tests.

Actually, it turns out we don't test the headers codegen there so the tests didn't fail.

skeqiqevian (approved):
LGTM

PR #607, opened 2024-03-30
Support Defining Globals for Instructions

SamirDroubi:


SamirDroubi:
It is expected for the avx2 ExoBLAS codegen tests to fail since this commit also minimized the number of empty lines generated in the source files.

src/exo/LoopIR.py
```diff
@@ -163,9 +166,12 @@ def __new__(cls, op):
                 fnarg*          args,
                 expr*           preds,
                 stmt*           body,
-                string?         instr,
+                instr?          instr,
                 srcinfo         srcinfo )
 
+    instr   = ( string          instr,
```
skeqiqevian, line 172: Writing `instr.instr` is a little confusing, how about `instr.c_instr` or something like that?

SamirDroubi: Yeah that's better. I will also rename `global_` to `c_global`.

SamirDroubi: Done

yamaguchi1024 (changes requested):
- you probably want a test which directly constructs `@instr(..)` with global includes
- also a test which includes some C stdlib headers

yamaguchi1024:
@SamirDroubi What's the status of this PR? Is this waiting for me or Kevin's review?

SamirDroubi:
@yamaguchi1024 yes

src/exo/API.py
```diff
@@ -341,7 +341,7 @@ def c_code_str(self):
         return decls + "\n" + defns
 
     def compile_c(self, directory: Path, filename: str):
-        compile_procs([self._loopir_proc], directory, f"{filename}.c", f"{filename}.h")
+        compile_procs([self], directory, f"{filename}.c", f"{filename}.h")
```
yamaguchi1024, line 344: I remember seeing this and we did talk about it but I forgot; was this a bug?

yamaguchi1024 (approved):

src/exo/API.py
SamirDroubi: It was untested code; someone changed the interface of compile_procs at some point and as a result this call was incorrect.

PR #608, opened 2024-03-31
Implement split_write scheduling operation

SamirDroubi:


src/exo/API_scheduling.py
```diff
@@ -1884,6 +1884,34 @@ def merge_writes(proc, block_cursor):
...
+            ->
+        `a = b`
+        `a += c`
+        ----------------------
+        `a += b + c`
+            ->
+        `a += b`
+        `a += c`
+        ----------------------
+
```
skeqiqevian, line 1909: When adding scheduling ops, I think we should try to specify forwarding rules in documentation, or at the very least, write a few test cases in `tests/test_cursors.py` so that we understand the behavior. I think for this scheduling operation, 
- cursor to the statement (and any subtree cursors) gets invalidated.
- blocks containing the statement will forward naturally.

src/exo/API_scheduling.py
```diff
@@ -1884,6 +1884,34 @@ def merge_writes(proc, block_cursor):
...
+    Split a reduce or assign statement with an addition on the RHS into two
+    writes.
+
+    This operation is the opposite of the last two cases of `merge_writes`.
+
+    args:
+        stmt    - cursor pointing to the assign/reduce statement.
+
+    rewrite:
+        `a = b + c`
```
skeqiqevian, line 1899: We might also want to support
```
a = c
a += b
```
In simple cases, the `b` vs `c` computation ordering doesn't matter, but in more advanced cases (e.g. `c` is a value that's pre-computable), we might want to support this behavior. Plus it's a pretty simple extension to add.

skeqiqevian (changes requested):
Suggested a quick extension to this scheduling operation + adding forwarding documentation.

src/exo/API_scheduling.py
SamirDroubi: You can call commute_expr on the rhs and then use this operation to get that behavior.

src/exo/API_scheduling.py
SamirDroubi: Yeah that's a good idea. I will add both the tests and documentation.

SamirDroubi: Done!

skeqiqevian (approved):
LGTM

PR #609, opened 2024-03-31
Simplify logical expressions containing True/False literals

SamirDroubi:


src/exo/LoopIR_scheduling.py
```diff
@@ -3307,16 +3310,28 @@ def map_binop(self, e: LoopIR.BinOp):
         elif e.op == "*":
             if is_const_zero(lhs) or is_const_zero(rhs):
                 return LoopIR.Const(0, lhs.type, lhs.srcinfo)
-            if isinstance(lhs, LoopIR.Const) and lhs.val == 1:
+            if is_const_val(lhs, 1):
```
skeqiqevian, line 3313: I'm surprised this additional simplification didn't change any golden test outputs. Do we simply not have tests for these cases? If so, we should add them.

skeqiqevian (approved):
LGTM. Was the motivation for this change the ability to leverage `if` statement conditions in simplification?

SamirDroubi:
> LGTM. Was the motivation for this change the ability to leverage `if` statement conditions in simplification?

Sometimes you start with procedures with a lot of conditions and as you specialize it, those conditions become always true or false. The Gemini code here that got fixed is one example. It is generally nice if those true/false disappeared because they clobber the code and to allow other simplifications to happen (like what happened in the Gemini code for example).

PR #613, opened 2024-04-06
Rewrite gemmini conv with user-level schedule ops

yamaguchi1024:
This PR rewrites the previous Gemmini conv artifact evaluation code to use user-level scheduling ops we defined for Gemmini matmul, such as `fission_as_much_as_possible` and `fuse_all_loops`. There are many TODOs to clean up the Gemmini code, such as 1) write a parameterized schedule that works for various resnet matmul and conv sizes (can we clean up `sched_one_block` and somehow clean up the loop reorganization code?), 2) run FireSim to check the performance. New schedules might achieve better performance than the previous one.


skeqiqevian (approved):
In terms of code reorganization, LGTM, but we should make sure to clean up overlap with `stdlib` later. 

We are not benchmarking the changes on FireSim for now because that is expensive.

Issue #318, opened 2023-02-02
Need to Add Overwrite Analysis to `stage_mem`


gilbo:
Right now, if you stage some memory in a statement that purely writes to the memory in question, it will not generate read or initialization code.  However, if there is some portion of the temporary buffer that is not written to by the statement, this could result in potentially overwriting some portion of the original buffer with garbage.

We thought we found a bug with this while talking with the Apple people.  We should check whether or not there is a bug and either (a) resolve this issue or (b) create a minimum failing test case before fixing the problem.

rachitnigam:
Let's try to write down a concrete example of this so we have a starting point. Seems related to the broader question making `stage_mem` more robust.

skeqiqevian:
```
def test_test():
    @proc
    def foo(x: i8[10]):
        for i in seq(0, 8):
            x[i] = 0.0

    foo = stage_mem(foo, foo.body(), "x[0:9]", "xtmp")
    print(foo)
```
incorrectly generates the following:
```
    xtmp: i8[9 - 0] @ DRAM
    for i in seq(0, 8):
        xtmp[i - 0] = 0.0
    for i0 in seq(0, 9 - 0):
        x[i0 + 0] = xtmp[i0]
```
which overwrites the value of x[8] with garbage data.

PR #614, opened 2024-04-06
Fix stage_mem logic for which expressions should use the staged memory

### LINKED ISSUE #446: Check_BufferRW in stage_mem is totally broken

yamaguchi1024:
- Only `stage_mem` accesses that always contained with the window expression. Don't `stage_mem` accesses that are always disjoint from the window expression. If there are accesses which are sometimes contained and sometimes disjoint, raise a SchedulingError. This allows a more precise analysis of read/write/reduced buffers.
- If all the replace of existing accesses fail, `stage_mem` should fail instead of allocating a new buffer unnecessarily.
- Deprecate Check_BufferRW

Closes #446


src/exo/LoopIR_scheduling.py
```diff
@@ -3613,6 +3615,49 @@ def get_typ_mem():
     else:
         new_typ = T.Tensor(shape, False, buf_typ.basetype())
 
+    def check_idx(idx, c, p):
+        assert len(idx) == len(w_exprs)
+        s = get_enclosing_stmt_cursor(c)._node
+        # check win.lo <= e.pt < win.hi
+        for i, w in zip(idx, w_exprs):
+            if not isinstance(w, tuple):
```
skeqiqevian, line 3623: This is currently buggy. We should also perform a check when `w` isn't a tuple. Consider the following test case: 
```
@proc
def foo():
    x: i8[10, 10, 10]
    for i in seq(0, 10):
        x[i, i, i] = 1.0

foo = stage_mem(foo, foo.find_loop("i"), "x[0:10, 0, 0:10]", "x_tmp")
```

src/exo/LoopIR_scheduling.py
```diff
@@ -3613,6 +3615,49 @@ def get_typ_mem():
...
+
+        return True
+
+    def check_win(idx, c, p):
+        assert len(idx) == len(w_exprs)
+        s = get_enclosing_stmt_cursor(c)._node
+
+        for i, w in zip(idx, w_exprs):
+            # check win.lo <= e.lo and e.hi <= win.hi
+            if not isinstance(w, tuple):
```
skeqiqevian, line 3641: Same here.

skeqiqevian (changes requested):

yamaguchi1024:
This PR unsurprising breaks the BLAS repo. @skeqiqevian Can you review the PR? @SamirDroubi Can we go through the changes to BLAS next week?

skeqiqevian:
The following should fail. In general, we should fail (instead of just "not replacing") when an existing window expr in the staged scope is not entirely a subset or not entirely disjoint with our staged window expr.

```
def test_counterexample():
    @proc
    def foo(N: size, A: i8[N, N]):
        sum: i8
        for i in seq(0, N):
            for jo in seq(0, N / 4):
                for ji in seq(0, 4):
                    A[i, 4*jo + ji] = 0.0
                    y = A[i:i+1, 4*jo+ji]
                    sum += y

    foo = stage_mem(foo, "for ji in _:_", "A[i, 4*jo:4*jo+4]", "tile")
    print(foo)
```

tests/golden/test_schedules/test_fail_stage_mem.txt
```diff
@@ -0,0 +1,15 @@
...
+            for k in seq(0, n / 4):
+                Btile: R[4, 4] @ DRAM
+                for i0 in seq(0, 4):
+                    for i1 in seq(0, 4):
+                        Btile[i0, i1] = B[i0 + 4 * i, i1 + 4 * k]
+                for ii in seq(0, 4):
+                    for jj in seq(0, 4):
+                        for kk in seq(0, 4):
+                            A[ii + 4 * i,
+                              jj + 4 * j] += Btile[ii, kk] * B[kk + 4 * k,
```
skeqiqevian, line 14: I think there's still a bug. This replace should fail because there are cases where the second read of `B` could overlap with the first read, e.g. if `i = j = k` and `ii = jj = kk`. In this particular test case, it's not fatal because `B` is read only, but if you add a `B[kk + 4 * k, jj + 4 * j] = 1.0` statement inside the loop, then `stage_mem` will break correctness.

tests/golden/test_schedules/test_stage_mem_okay.txt
```diff
@@ -0,0 +1,9 @@
+def foo(x: i8[10, 10, 10] @ DRAM):
+    y: i8 @ DRAM
+    x_tmp: i8[10, 10] @ DRAM
+    for i0 in seq(0, 10):
+        for i1 in seq(0, 10):
+            x_tmp[i0, i1] = x[i0, 0, i1]
+    for i in seq(0, 10):
+        x[i, i, i] = 1.0
```
skeqiqevian, line 8: This also seems wrong. `x[i, i, i]` overlaps with `x[0:10, 0, 0:10]` when `i = 0`. This particular code example happens be okay, but if you replaced `x[2, 0, 3]` with `x[0, 0, 0]`, then `stage_mem` would break correctness.

skeqiqevian (changes requested):
I didn't look at all the tests, but the two that I looked at seem incorrect. I think more careful index analysis is still required.

tests/golden/test_schedules/test_stage_mem_okay.txt
yamaguchi1024: Ugh, thanks for noticing this!

yamaguchi1024:
I'm giving up, assigning @skeqiqevian 🙏

skeqiqevian:
As currently implemented, this will not close #318.

skeqiqevian:
@yamaguchi1024 can you review when you get a chance? 

I removed the changes about writes shadowing reads since that implementation wasn't correct. In general, you have to reason about the entire block's read and write sets, which seems hard to do. I think doing such reasoning might be able to address #318 if we check that the block's write set is the same as the staged window.

yamaguchi1024:
@skeqiqevian Thanks Kevin! I agree that we have to reason about the entire block's read and write sets to make it correct, but I'm not sure why you had to remove it. The incorrect write-shadowing-read behaviour is not introduced in this PR. That's why BLAS CI is failing.

w/o this PR:
```
void exo_dasum_stride_1( void *ctxt, int_fast32_t n, struct exo_win_1f64c x, double* result ) { 
// assert stride(x, 0) == 1
double result_;
result_ = 0.0;
__m256d var10;
var10 = _mm256_setzero_pd();
__m256d var15[8];
var15[0] = _mm256_setzero_pd();
```

w/ this PR:
```
5: void exo_dasum_stride_1( void *ctxt, int_fast32_t n, struct exo_win_1f64c x, double* result ) {         
5: // assert stride(x, 0) == 1                    
5: double result_;                                                                                         
5: result_ = *result;                                                                                      
5: result_ = 0.0;                                                                                          
5: __m256d var10;                                                                                                                                                                                                     
5: var10 = _mm256_setzero_pd();                                                                            
5: __m256d var15[8];                                                                                       
5: var15[0] = _mm256_setzero_pd();   
```
We have this extra `result_ = *result;`.

Even if reasoning about complete write-shadowing-read is out-of-scope of this PR, can't we at least make a conservative but sound shadowing effect? I think what I was trying to do was incorrect because it didn't look at the buffer indices of read and writes.

skeqiqevian:
I couldn't think of a tractable way to solve your problem - you need to check that every read is shadowed by a write in order to make that change. And even if there is a simple, but conservative check, I think it makes more sense to include that in separate PR.

yamaguchi1024:
Okay I think that's fine. I want to approve this PR but I can't because I started it. Do you want to approve it?

Also @SamirDroubi heads up this will change BLAS codegen slightly, but I think it's a negligible overhead

SamirDroubi:
> Okay I think that's fine. I want to approve this PR but I can't because I started it. Do you want to approve it?
> 
> Also @SamirDroubi heads up this will change BLAS codegen slightly, but I think it's a negligible overhead

I would like first to see what changed and make sure it is not a big deal. I will build BLAS with this tomorrow and see the diff.

skeqiqevian (approved):
Yuka approved my changes above.

PR #618, opened 2024-04-17
Multistage programming

kehemo:


kehemo:
Python is the metalanguage and the Exo base language has not been changed in any breaking way

Python unquote blocks are evaluated using exec. Each unquote behaves like a new scope in python. Both unquote and quote blocks have parent scopes that look at the nearest enclosing block of the same type. For example:
```
@proc
def foo():
    a: i8
    with unquote:
        b = 1
        with quote:
            a = 1
            with unquote:
                print(b)
```
When this procedure is evaluated, it will print "1", and then result in a procedure that is equivalent to:
```
@proc
def foo():
    a: i8
    a = 1
```

kehemo:
Some relatively minor things are still broken:
- I make the assumption that free variables in Python cannot be mutated, but I later realized this is not true due to the `nonlocal` keyword.
- I do not check the AST when mangling some identifiers that I splice into the code. I only check variables from enclosing scope in that case
- I allow users to write return statements inside unquote blocks, which kinda doesn't make sense

kehemo:
This branch is only a subset of the desired features for metaprogramming. In future PRs, consider:
- adding an API for reflection on Exo statement objects and expressions
- adding macros (functions outside of exo procs that can be used in unquoted blocks)
- adding syntax that lets users create an Exo identifier and unquote it in any position an Exo identifier would appear

yamaguchi1024:
Thanks Kenneth! I will review this PR more carefully, but meanwhile can you write a documentation for metaprogramming features and limitations? (see https://github.com/exo-lang/exo/pull/729 for sample documentation for other features).

tests/test_typecheck.py
```diff
@@ -80,14 +80,14 @@ def foo(n: size, A: R[n] @ GEMM_SCRATCH):
...
-    def sin(x: f32):
+    def sin_proc(x: f32):
         y: f32
         y = sin(x)
 
 
 def test_sin2():
     @proc
-    def sin(x: f32):
+    def sin_proc(x: f32):
```
yamaguchi1024, line 90: Was this change necessary? Is it because the procedure name is conflicting with the extern name? 
I think procedure definitions should be able to overwrite whatever other global and local variables out there

tests/test_metaprogramming.py
```diff
@@ -0,0 +1,378 @@
...
+    def foo(a: i32):
+        with python:
+
+            def bar(x):
+                return ~{x + 1}
+
+            with exo:
+                a = {bar(~{a + 1})}
+
+    c_file, _ = compile_procs_to_strings([foo], "test.h")
```
yamaguchi1024, line 320: Why are you compiling the proc into C for every test? I think you should also sanity-check the ExoIR printing.
Can modify the asserts to sth like:
```
assert str(foo) + c_file == golden
```

yamaguchi1024: Also, you know that you can update all the golden tests by running `pytest tests/test_metaprogramming.py --update-golden` , right?

src/exo/frontend/pyparser.py
```diff
@@ -15,6 +15,9 @@
 from ..core.prelude import *
 from ..core.extern import Extern
 
+from typing import Any, Callable, Union, NoReturn, Optional
+import copy
+from dataclasses import dataclass
```
yamaguchi1024, line 20: What are those imports? Are they necessary? If so, you should add them as dependencies to `requirements.txt`

src/exo/frontend/pyparser.py
```diff
@@ -35,54 +38,106 @@ def str_to_mem(name):
     return getattr(sys.modules[__name__], name)
 
 
+@dataclass
+class SourceInfo:
```
yamaguchi1024, line 42: Why did you need a wrapper for `SrcInfo`?

src/exo/frontend/pyparser.py
```diff
@@ -831,12 +1307,79 @@ def parse_array_indexing(self, node):
...
+            for e in dims:
+                if sys.version_info[:3] >= (3, 9):
+                    srcinfo = self.getsrcinfo(e)
+                else:
+                    if isinstance(e, pyast.Index):
+                        e = e.value
+                        srcinfo = self.getsrcinfo(e)
+                    else:
+                        srcinfo = self.getsrcinfo(node)
+                if isinstance(e, pyast.Slice):
```
yamaguchi1024, line 1354: Can you add tests to cover the lines Codecov is reporting?

src/exo/frontend/pyparser.py
```diff
@@ -105,30 +160,339 @@ def pattern(s, filename=None, lineno=None, srclocals=None, srcglobals=None):
...
+    _inner: Any  # note: strict typing is not possible as long as PAST/UAST grammar definition is not static
+
+
+@dataclass
+class ExoStatementList:
+    _inner: tuple[Any, ...]
+
+
+@dataclass
+class QuoteReplacer(pyast.NodeTransformer):
```
yamaguchi1024, line 201: You should definitely add docstrings to each Class and functions you defined

tests/test_typecheck.py
kehemo: The change is necessary because the sin function call inside the proc is parsed as an implicit unquote in the metalanguage, which means it is looked up like a normal python variable named sin would be looked up at that point. But, if the proc is named sin, it is registered as a local variable in the scope which shadows the sin extern that is imported as a global variable.

src/exo/frontend/pyparser.py
kehemo: The imports are parts of the python standard library just to give some type annotations and reduce some clutter

src/exo/frontend/pyparser.py
kehemo: I decided to make a wrapper because before we just passed a getsrcinfo function into Parser, which I thought was confusing because then you don't know where the behavior of getsrcinfo is defined. I think it is neater if it is wrapped in an object with a defined get_src_info method that you can jump to.

tests/test_typecheck.py
yamaguchi1024: Hmmmm I see what you mean but this seems quite cryptic.... It's way too opaque to users... How bad is it going to be if we don't support implicit unquotes?

kehemo: I think the code becomes pretty cluttered if there are no implicit unquotes because you will have to put braces around every variable you reference from the metalanguage.

Also, I think it should be expected that if you name a proc "sin" it will shadow any other "sin" inside its scope. In most other languages, you can refer to the function you are currently inside by its name (e.g. recursion), so it doesn't seem very surprising.

kehemo: As a side note, if you try compiling this exo:
```
@proc
def sin(x: f32):
    x = sin(x)
```
the result is:
```
#include <math.h>
// sin(
//     x : f32 @DRAM
// )
void sin( void *ctxt, float* x ) {
*x = sin((float)*x);
}
```
which is actually invalid C because the `sin` inside the function will refer to the proc `sin` not the function from `math.h`.
```

Comment (yamaguchi1024) [reply tests/test_typecheck.py:90]: I see okay, this makes sense. Can you add a test case of declaring a proc named `sin` that shadows the extern sin, and call the proc `sin` from another procedure `foo`, and see if that works as expected?

File: tests/test_metaprogramming.py (hunk 2 above)
Comment (kehemo) [reply tests/test_metaprogramming.py:324]: complete

File: tests/test_typecheck.py (hunk 1 above)
Comment (kehemo) [reply tests/test_typecheck.py:90]: complete

File: src/exo/frontend/pyparser.py [hunk 7]
```diff
@@ -493,16 +994,70 @@ def parse_num_type(self, node, is_arg=False):
...
             )
         else:
-            self.err(node, "unrecognized type: " + pyast.dump(node))
+            unquote_eval_result = self.try_eval_unquote(node)
+            if len(unquote_eval_result) == 1:
+                unquoted = unquote_eval_result[0]
+                if isinstance(unquoted, str) and unquoted in _prim_types:
+                    return _prim_types[unquoted]
+                else:
+                    self.err(node, "Unquote computation did not yield valid type")
```
yamaguchi1024, line 1003: We need else: on `if len(unquote_eval_result) == 1:` and raise error

src/exo/frontend/pyparser.py
```diff
@@ -203,30 +647,90 @@ def __init__(
...
+                        self.make_exo_var_asts(self.getsrcinfo(unquote_node)),
+                    ).interpret_unquote_expr(unquote_node),
+                )
+                if unquote_node.id in cur_locals or unquote_node.id in cur_globals
+                else tuple()
+            )
+        else:
+            return tuple()
+
     def eval_expr(self, expr):
```
yamaguchi1024, line 724: Can you add tests for
```
mems=[DRAM, ...]
x : f32 @ mems[0]
```

kehemo:
admittedly the docs could use better motivated examples, but I found it difficult to include any without overly complicating what should also serve as a reference that is easily skimmable. I think the first limitation (lack of quote outside of exo procs) also holds metaprogramming back by a lot more than i originally anticipated, but we should try to merge the existing features first and then hopefully reduce the limitations later.

I considered documenting how metaprogramming works internally in pyparser, but besides my existing comments, it doesn't really seem like there is a space for a more in-depth explanation.

yamaguchi1024 (approved):

PR #628, opened 2024-04-25
Allow divide_dim to divide to non-literal expressions

SamirDroubi:


src/exo/API_scheduling.py
```diff
@@ -1372,8 +1372,6 @@ def divide_dim(proc, alloc_cursor, dim_idx, quotient):
         `x : R[n, 3, 4, m]`
         `x[i, j / 4, j % 4, k] = ...`
     """
-    if quotient == 1:
```
yamaguchi1024, line 1375: Can you change the docstring above "This limited implementation of `divide_dim` requires that the dimension being divided is constant itself." ?

src/exo/LoopIR_scheduling.py
```diff
@@ -728,25 +768,10 @@ def ceildiv(lhs, rhs):
     elif tail_strategy in ["cut", "cut_and_guard"]:
         outer_hi = szop("/", N, inner_hi)  # floor div
     elif tail_strategy == "perfect":
-        if not isinstance(N, LoopIR.Const):
```
yamaguchi1024, line 731: This is great!

src/exo/LoopIR_scheduling.py
```diff
@@ -303,6 +321,28 @@ def move_back(c):
...
+# --------------------------------------------------------------------------- #
+# IR Building Helpers
+
+
+def divide_expr(e, quot):
+    assert isinstance(e, LoopIR.expr)
+    if isinstance(quot, int):
+        quot_int = quot
+        quot_ir = LoopIR.Const(quot, e.type, e.srcinfo)
+    elif isinstance(quot, LoopIR.expr):
```
yamaguchi1024, line 334: This should be `elif isinstance(quot, LoopIR.Const):`

yamaguchi1024 (changes requested):
lgtm as long as the docstring and the `elif isinstance(quot, LoopIR.Const):` thing is fixed

src/exo/API_scheduling.py
SamirDroubi: Done!

src/exo/LoopIR_scheduling.py
SamirDroubi: Good catch!

yamaguchi1024 (approved):

PR #630, opened 2024-04-27
Update dependencies and requirements

yamaguchi1024:
- Update dependencies
- Remove jupyter, pillow, and numpy from requirements
- Update tox.ini
- Remove flake files


.github/workflows/gemmini.yml
```diff
@@ -25,6 +25,7 @@ jobs:
           apt-get remove -y cmake
           apt-get install -y ninja-build
           python -m pip install -r requirements.txt
+          python -m pip install -r dev-requirements.txt
```
SamirDroubi, line 28: I recommend splitting the requirements into two files: `exo-dev-requirements.txt` (for people who wants to develop exo) and `exo-requirements` for people who just want to use exo. Then, you can include them both from a `requirements.txt` file using the `-r <file>` format: https://pip.pypa.io/en/stable/reference/requirements-file-format/#referring-to-other-requirements-files

This way any one person (dev or just user) need to run one command to install everything and also dependabot can discover all the requirements (it looks only for `requirements.txt`).

SamirDroubi (changes requested):

requirements.txt
```diff
@@ -1,17 +1,6 @@
...
-coverage==6.3.2
-jupyter-contrib-nbextensions==0.5.1
-numpy==1.23.4
-pre-commit==3.6.0
-pytest-cov==3.0.0
-pytest-xdist==3.0.2
-pytest==7.1.1
-tox==3.24.5
-yapf==0.32.0
 z3-solver==4.12.6.0
```
SamirDroubi, line 5: Can we bump z3 to 4.13.0.0? I am actually not sure why dependabot didn't do that...

SamirDroubi (changes requested):

.github/workflows/gemmini.yml
yamaguchi1024: I think it makes sense to have requirements.txt just for users, instead of including the dev dependencies. CI will look at tox.ini for test requirements separately, so I don't think that's an issue?

requirements.txt
yamaguchi1024: Done!

.github/workflows/gemmini.yml
SamirDroubi: Yeah but dependabot would then not see the dev requirements? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pip-and-pip-compile

yamaguchi1024: From this discussion ( https://github.com/dependabot/dependabot-core/issues/3940#issuecomment-1227005267 ) and dependabot's source code ( https://github.com/dependabot/dependabot-core/blob/main/python/helpers/lib/parser.py#L84 ) it seems like they're scanning all *.txt files and is not relying on "requirements.txt" pattern match, so I think having two separate txt files is fine

SamirDroubi: Ah I see. That's not clear from the documentation. Good to know!

SamirDroubi (approved):

PR #637, opened 2024-04-30
Externalize builtins and fix bugs

### LINKED ISSUE #505: proc/builtin calls semantics
I don't think there is very well defiend semantics about subprocedure calls or builtin calls. We need those to be well-defined and documented somewehere.

This may speak to a bigger problem where we don't track anywhere the syntax and semantics of Exo programs which may imply that we don't have accurate expectations of what a rewrites, anylsis, and codegen can/cannot do.

Some examples, why this is confusing:

1) This code is accepted by the front-end and rejected by codegen because we are passing arguments that are reads with an index:
```
@proc
    def select2(x: f32[4], result: f32):
        result = select(x[0], x[1], x[2], x[3])
```
2) This code accepted by both front-end and codegen:
```
@proc
    def dot(m: size, x: f32[m], y: f32[m], r: f32):
        r = 0.0
        for i in seq(0, m):
            r += x[i] * y[i]

@proc
    def hoge(n: size, x: f32[n], y: f32[n]):
        xy: f32
        dot(n, x, y, xy)
```
There is also a somewhat confusing representation internally in example (2) where the second argument `x` is passed as a LoopIR.Read with idx = [] to imply that this is passing the entire tensor.

yamaguchi1024:
- Rename builtins to _externs_
- Externalize _externs_
- Fix bug 1: extern did not work on read expressions with index accesses
- Fix bug 2: extern fnarg codegen should be `comp_e` not `comp_fnarg`
- Fix bug 3: extern did not work on constants
- Change pattern_match and parse_fragment to take local and global variables in scope.

Fixes #505


src/exo/LoopIR_compiler.py
```diff
@@ -434,17 +434,17 @@ def from_lines(x):
 // Compiler feature macros adapted from Hedley (public domain)
 // https://github.com/nemequ/hedley
 
-#if defined(__has_builtin)
```
SamirDroubi, line 437: I am confused. Why were these changed? These are referring to C builtins, not Exo.

SamirDroubi (changes requested):

src/exo/API_cursors.py
```diff
@@ -72,9 +72,9 @@ class Cursor(ABC):
                | Literal( value : bool, int, or float )
                | UnaryMinus( arg : Expr )
                | BinaryOp( op : str, lhs : Expr, rhs : Expr )
-               | BuiltIn( name : str, args : ExprList )
+               | Extern( name : str, args : ExprList )
                | WindowExpr( name : str, idx : *(see below) )
-               | BuiltIn( name : str, args : ExprList )
```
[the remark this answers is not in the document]
yamaguchi1024: lol yeah good catch

src/exo/LoopIR_compiler.py
```diff
@@ -434,17 +434,17 @@ def from_lines(x):
 // Compiler feature macros adapted from Hedley (public domain)
 // https://github.com/nemequ/hedley
 
-#if defined(__has_extern)
-#  define EXO_HAS_extern(extern) __has_extern(extern)
+#if defined(__has_builtin)
```
SamirDroubi, line 437: All these macros should be reverted back to what they used to be. The reason they exist is orthogonal to your PR.

yamaguchi1024: Yeah was doing that

src/exo/LoopIR_compiler.py
```diff
@@ -375,11 +375,10 @@ def from_lines(x):
     public_fwd_decls = []
 
     # Body contents
-    memory_code = _compile_memories(find_all_mems(proc_list))
```
SamirDroubi, line 378: I presume you made this change so that if a builtin is onlu used within the definition of an instruction, then we don't need to generate its "globals". I think I can agree with that, however, this might cause problems for memories e.g.

@instr(...)
def foo(...):
     a: var @ SOME_NEW_MEMORY
     # Do something with `a`

Now, to be fair, I don't know of any case where you really need to do this, but it legal Exo code and we should support it.

src/exo/LoopIR_compiler.py
```diff
@@ -375,11 +375,10 @@ def from_lines(x):
...
-    builtin_code = _compile_builtins(find_all_builtins(proc_list))
     private_fwd_decls = []
     proc_bodies = []
     instrs_global = []
+    new_proc_list = []
```
SamirDroubi, line 381: I found this name confusing. Maybe you can call it non_instr_procs. Or at least add a comment after it.

src/exo/extern.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, name):
+        self._name = name
+
+    def name(self):
+        return self._name
+
+    def globl(self, prim_type):
+        raise NotImplementedError()
+
+    def typecheck(self, args):
```
SamirDroubi, line 29: I think it is confusing to ask the user to implement type checking. If all we need to do is check the number of arguments, check that all of them are reals, and I think we want to check that all of them are the same precision. We can check all of these things on our own in the frontend and at codegen. You probably want to check all of these in the frontend. Ast codegen, you can recheck all of them as a sanity check, but I think the last condition is the only one that could be invalidated through rewrites.

src/exo/API.py
```diff
@@ -267,7 +267,7 @@ def body(self):
         block = self._root()._child_block("body")
         return C.lift_cursor(block, self)
 
-    def find(self, pattern, many=False):
+    def find(self, pattern, many=False, call_depth=1):
```
SamirDroubi, line 270: I don't think all these `call_depth` are necessary at the API level. It seems like you needed to change the value (internally) in one place and you added them to all `find` functions. I couldn't find where you are using this from user-code. I would revert most of these changes and make the change in the exact place where you needed it (within the compiler implementation).

yamaguchi1024: call_depth is necessary because we need to know how many stack frames `match_pattern` has when we get there, because we inspect the stack frame to get globals and locals. Internally, proc.find (this function) is calling into cursor.find, and both APIs are exposed to the user. More confusingly, cursor.find is calling into find (in API_cursors.py). So the problem is that multiple APIs are exposed to the user, and we have to keep track of call_depth accordingly.

src/exo/LoopIR_compiler.py
yamaguchi1024: This change was made because we have to run precision analysis on the proc_list before calling `_compile_builtins`. But yeah, it might be confusing not to generate globls for instructions. I'll change that.

src/exo/LoopIR_compiler.py
yamaguchi1024: `analyzed_proc_list` ?

src/exo/libs/externs.py
```diff
@@ -0,0 +1,248 @@
...
+        if not atyp.is_real_scalar():
+            raise _EErr(
+                f"expected argument 1 to be a real scalar value, but "
+                f"got type {atyp}"
+            )
+        return atyp
+
+    def globl(self, prim_type):
+        s = (
+            f"{prim_type} _relu_{prim_type}({prim_type} x) " + "{\n"
```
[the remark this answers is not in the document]
yamaguchi1024: There's no function overloading in C, right? We can consider using `_Generics`, but we'd have to really on C11 standard for codegen which sounds like a bad idea

src/exo/extern.py
yamaguchi1024: All the examples here happen to have the same arguments, but that's not always the case. I think it's reasonable to use externs to implement a user-defined casting (e.g., cast by clamping instead of our default C casting). So, I think typecheck should be exposed to users, not to constrain the kind of function they want to write.

src/exo/API.py
SamirDroubi: I think if you scroll through all your modifications you will find that all the uses cases where you specify `call_depth` can be statically determined and don't need to be exposed as an argument to the user. The issue is that some API functions work by calling other API functions and you need to change the `call_depth` in one of those cases from the caller to account for the extra call. You can also have some internal (private) `find` function that takes in `call_depth` and then you don't have surface this to the user code.

One other thing you could do is to refactor the logic that is inspecting the user stack here: https://github.com/exo-lang/exo/blob/91fa03bac8ddb30ea4cd4835114f1a8bea026544/src/exo/pattern_match.py#L82 and resolve at the API-level call. Then, you just pass in that information downstream instead of maintaining `call_depth` across compiler code which is somewhat confusing.

yamaguchi1024: > You can also have some internal (private) find function that takes in call_depth and then you don't have surface this to the user code.

I'm not sure what this means? Can you explain a bit more?

> refactor the logic that is inspecting the user stack here:

That would work, but then we'll end up passing around caller frameinfo across the compiler which is doubly confusing. It'll be sth like
```
    def find(self, pattern, many=False, frameinfo=None):   
        if frameinfo == None:
             frameinfo = ...                                                                                                                                                             
        return find(self._impl, self._proc, pattern, many, frameinfo=frameinfo) 
```
in `class Cursor`, which I think is worse

Also I think we do need to expose call_depth to users, because when you define a composed function in user library, you need to change the call_depth so that the stack frame points to the actual user code, not the library implementation definition.

Imagine you have `find_user_library_wrapper` for some reason in your library.
```
def find_user_library_wrapper(proc, cursor, ...):
    ... # do sth
    return proc.find(cursor, ...., call_depth=2)
```
call_depth in this case has to be 2

SamirDroubi: With regard to the internal function, I am saying that you needed to add the `call_depth` to API-level functions because you had member API functions calling other member API functions. Instead, you can have them all call some internal non-API member function `_find` that assumes that it is always at depth = 1 because it is being called from an API function. A different way is to not make the API functions call each other.

yamaguchi1024: I see sure that'll work, but I think that'll give an incorrect stack frame for the `find_user_library_wrapper` example above

SamirDroubi: I don't think you need that for user defined functions. You need it now because we don't have a good way to resolve patterns at the call-level, but we resolve them when we call `find` methods within the implementation. Once we add that feature this wouldn't be necessary.

Currently:
```
def user-op(proc, expr):
      e= proc.find(expr) <- resolved here
```

What we will have in the future:
```
@sched-op
def user-op(proc, expr: ExprCursor):
      # expr is already resolved before we enter the function and `call-depth` is a deterministic static value. 
```

yamaguchi1024: Yeah with `@sched-op` I believe it'll work, but that's requires more engineering and I'm not even sure what's the roadmap for that is.

src/exo/extern.py
SamirDroubi: The problem with this is that it contradicts the fact that you cannot do mixed precision operations. You might want to be able to specify the return type to be something else for the case you are talking about, but I still think the parameters should be of the same type. This is especially the case when you specify `def globl(self, prim_type)` is associated with a single type.

yamaguchi1024: I think it's totally fine to be able to mix precisions (not in the externs examples we have right now, but why not?). To define a user-defined casting with externs, the argument type and the return type needs to be different.

src/exo/API.py
SamirDroubi: I have traced through the code and I don't think srcinfo is used in any meaningful way down this path anyways since we are not constructing an IR node. If you fail to find something, you will get something like:
```
exo.new_eff.SchedulingError: find: failed to find matches
E           Pattern: z[_] = _
```
And you can look through the python stack trace to find where the find failed. We currently don't use that information to report any such errors. I am not sure why it was important to pass the `call_depth` information to begin with.

src/exo/extern.py
SamirDroubi: Here is an example where this becomes problematic:

```
@proc
    def foo(a: f32, b: f64, c: i32, d: i8, e: f64):
      e = select(a, b, c, d)
```

which will generate the following on the code from this PR:
```
int8_t _select_int8_t(int8_t x,int8_t v,int8_t y,int8_t z) {
    if (x < v) return y;
    else return z;
}

// foo(
//     a : f32 @DRAM,
//     b : f64 @DRAM,
//     c : i32 @DRAM,
//     d : i8 @DRAM,
//     e : f64 @DRAM
// )
void foo( void *ctxt, const float* a, const double* b, const int32_t* c, const int8_t* d, double* e ) {
*e = (double)(_select_int8_t((int8_t)*a, (int8_t)*b, (int8_t)*c, (int8_t)*d));
}
```
This would compute the wrong thing.

Previously, the implementation implemented `select` with double as the type of all the arguments and casted any argument into that which I thought made sense because it is a best-effort to have an implementation that always has a high-enough precision. Yet, that implementation could have also broken in some cases.

This is why I think unifying the types of the arguments makes sense because we have always said "Exo requires all data-expressions to have consistent precision, (e.g. multiplying an f32 and i8 is forbidden) but inserts type-casts as necessary just before writing or reducing data values." Now, you can generate those extern implementations correctly since there is exactly one type.

src/exo/API.py
yamaguchi1024: call_depth is used to get locals and globals in `src/exo/pattern_match.py`. Yeah we don't really good make use of srcinfo.
```
   func_locals = ChainMap(stack_frames[call_depth].frame.f_locals)
   func_globals = ChainMap(stack_frames[call_depth].frame.f_globals)
``

File: src/exo/extern.py (hunk 6 above)
Comment (yamaguchi1024) [reply src/exo/extern.py:29]: So I agree that `select`, should, check the type consistency of arguments, but my argument is that there's no reason to impose that constraint to users. Users should be able to write clamp casting with externs.

Comment (yamaguchi1024) [reply src/exo/extern.py:29]:
```

class _Clamp(Extern):
    def __init__(self):
        super().__init__("clamp")

    def typecheck(self, args):
        if len(args) != 1:
            raise _EErr(f"expected 1 argument, got {len(args)}")

        atyp = args[0].type
        if not isinstance(atyp, T.f32):
            raise _EErr(f"expected argument 1 to be f32")
        return atyp

    def globl(self, prim_type):
        assert prim_type == "float"
        s = (
            f"i8 _clamp_(float x) " + "{\n"
            "    ...."
            "}\n"
        )
        return s

    def compile(self, args, prim_type):
        assert prim_type == "float"
        return f"_clamp_({args[0]})"

clamp = _Clamp()
```

Comment (SamirDroubi) [reply src/exo/extern.py:29]: This example still doesn't require mixed precision arguments (the arguments are all of the same type, the return value 
 is different). In any case, if you really think type-checking should be something that is exposed for the users. I would suggest rewrite this class such that it defines a signature of the arguments and return types e.g. get_signature(): return `(ExoType.i8, (ExoType.f32, ))` and we automatically generate a type check based on this i.e. Tuple(ReturnType, Tuple(Arg0Type, Arg1Type, ...)).

**Comment by SamirDroubi** [comment]
I think a more ergonomic way of implementing this feature would be something like the following:
```
@extern("{a_data} < {b_data} ? {c_data} : {d_data}")
def select(a: f32, b: f32, c: f32, d: f32) -> f32:
      return c if a < b else d
```
This would effectively encode all the information you are encoding in the class and it is more akin to how we externalized instructions definitions. There is the issue however of how to make sure these externs' arguments are dynamically typed which was the case before. We might want to allow these externs to be overloaded to avoid having to write on per-type or maybe this is not an issue because it is the same problem for instruction. We might though need to allow them to be rewritten modulo precisions. e.g. `swap_builtin(proc, proc.find("select_R(_)"), select_f32)` or something like that.

**Comment by yamaguchi1024** [comment]
I like this interface!! It's like we're adding Exo type annotation (e.g., f32) to Python function, but the `@extern` function body is in Python.

**Comment by yamaguchi1024** [comment]
@SamirDroubi I think `@extern` makes sense and we should do it at some point, but I don't have an energy to do it now so I'll just move the typechecking into the compiler and parse the proc's externs instead of using call_depth

**Comment by yamaguchi1024** [comment]
Flagging here that we don't have documentation for builtins/externs. Maybe worth fixing that in this PR

File: src/exo/libs/externs.py [hunk 9]
```diff
@@ -0,0 +1,248 @@
...
+
+        atyp = args[0].type
+        if not atyp.is_real_scalar():
+            raise _EErr(
+                f"expected argument 1 to be a real scalar value, but "
+                f"got type {atyp}"
+            )
+        return atyp
+
+    def globl(self, prim_type):
```
yamaguchi1024, line 209: FIXME: sigmoid is not in math.h!

src/exo/pyparser.py
```diff
@@ -166,15 +172,18 @@ def __init__(
         self.is_fragment = is_fragment
 
         self.push()
+        special_cases = ["stride"]
```
akeley98, line 175: Creating this special_cases list and then searching it linearly (in the `s.value.func.id in special_cases`) looks like potentially expensive "accumulated cost" considering how many times this parser is used, e.g., to parse all procs and all  patterns in `@instr`. Can this be an `is_special_case` lambda that just searches for the func id in the 2 namespaces (+ the special stride case)?

akeley98 (changes requested):
I tried to scan through all the nontrivially changed files and the grammar changes and it looks reasonable so far, except the one performance note I made. I don't understand what call_depth is doing very well ... this is something about being able to look-up Python local variables?

src/exo/pyparser.py
akeley98: Something like

```
        def is_special_case(func_id):
            return (
                func_id == "stride"
                or isinstance(self.globals.get(func_id), Extern)
                or isinstance(self.locals.get(func_id), Extern)
            )
...
            if len(module_ast) == 1:
                s = module_ast[0]
                if isinstance(s, pyast.Expr) and (
                    not isinstance(s.value, pyast.Call)
                    or is_special_case(s.value.func.id)
                ):
                    is_expr = True
```

src/exo/prec_analysis.py
```diff
@@ -224,6 +246,18 @@ def coerce_e(self, e, btyp):
             assert rhs.type == btyp
 
             return LoopIR.BinOp(e.op, lhs, rhs, btyp, e.srcinfo)
+        elif isinstance(e, LoopIR.Extern):
+            assert e.type == T.R
+            # coerce if T.R
+            args = []
+            for a in e.args:
+                if a.type == T.R:
+                    args.append(self.coerce_e(a, btyp))
```
kehemo, line 255: are externs expected to support every available precision on Exo?

src/exo/pyparser.py
kehemo: i think `special_cases` is only used once which is to check patterns that are of the form `<extern>(...)` so I don't think there is a big cost to doing this

kehemo:
> I tried to scan through all the nontrivially changed files and the grammar changes and it looks reasonable so far, except the one performance note I made. I don't understand what call_depth is doing very well ... this is something about being able to look-up Python local variables?

yeah, when the user calls the api to create a pattern, the pattern might reference externs in the local scope or global scope where the user did the call. the api might make additional calls to other api functions, which then need to figure out the local scope of the function that originally called into the api. so the api needs to keep track of how many calls were made after the user initially called into the api

kehemo (commented):
looks good

kehemo (approved):

src/exo/prec_analysis.py
yamaguchi1024: Yes! But this is coercing the argument when it was T.R (which means the precision was not specified during scheduling)

src/exo/pyparser.py
yamaguchi1024: I think this is fine

PR #645, opened 2024-05-03
Add restrict qualifier to pointers at codegen

SamirDroubi:


yamaguchi1024:
Is this really safe? IIUC there's some code (like syr) that actually aliases the call to Exo kernels. Also, did you see any speedup from changing this?

skeqiqevian:
What was the reason we needed aliasing for `syr` again? Is it so that you can perform `stage_mem` correctly on only one of the vectors without impacting the other?

yamaguchi1024 (approved):
LGTM, but you should remove the aliasing kernel call after the stage_mem PR gets mergeed

yamaguchi1024:
Closing due to inactivity

PR #651, opened 2024-05-28
Prevent incomplete indices from pattern matching

skeqiqevian:
I'm not sure if this was previously an intentional choice, but it feels weird that something like `x[i]` can successfully match with `x[i, j]`.


skeqiqevian:
Turned out to be a bigger can of worms than I thought, so I'm going to close it.

Issue #652, opened 2024-05-29
Unknown Z3 error from remove_loop


yamaguchi1024:
Object code:
```
def matmul_on_gemmini(N: size, M: size, scale: f32 @ DRAM, act: bool @ DRAM,                                              
                      A: i8[N, 512] @ DRAM, B: i8[512, M] @ DRAM,                                                                                                                                                                                    
                      C: i8[N, M] @ DRAM):                                                                                
    assert N % 256 == 0                                                                                                   
    assert M % 256 == 0                                                                                                                                                                                                                              
    config_st_acc_i8(scale, stride(C, 0), act)                                                                            
    config_matmul()                                                                                                       
    config_ld_i8_id1(stride(A, 0))             
    config_ld_i8_id2(stride(B, 0))               
    config_zero()                                                                                                         
    A_tmp: i8[16, 32, 16, 16] @ GEMM_SCRATCH      
    B_tmp: i8[32, 16, 16, 16] @ GEMM_SCRATCH                                                                              
    res: i32[16, 16, 16] @ GEMM_ACCUM                      
    for ioo in seq(0, N / 256):                                                                                           
        for joo in seq(0, M / 256):                         
            for ioi in seq(0, 16):                                                                                        
                for joio in seq(0, 4):                                                                                    
                    for joii in seq(0, 4):                                                                                
                        do_zero_acc_i32(16, 16, res[joii + 4 * joio, 0:16,                                                
                                                    0:16])
    for ioo in seq(0, N / 256):      
        for joo in seq(0, M / 256):                                                                                       
            for ioi in seq(0, 16):                                                                                        
                for joio in seq(0, 4):                                                                                    
                    for koo in seq(0, 8):                                                                                 
                        if joio == 0:           
                            do_ld_i8_block_id1(              
                                16, 4, A[16 * ioi + 256 * ioo:16 + 16 * ioi +                                                                                                                                                                        
                                         256 * ioo, 64 * koo:64 + 64 * koo],                                                                                                                                                                         
                                A_tmp[ioi, 4 * koo:4 + 4 * koo, 0:16, 0:16])
        for joo in seq(0, M / 256):                                                                                                                                                                                                                  
            for ioi in seq(0, 16):                                                                                        
                for koo in seq(0, 8):                                                                                     
                    for koi in seq(0, 4):                                                                                 
                        for joio in seq(0, 4):                                                                            
                            if ioi == 0:                                                                                  
                                do_ld_i8_block_id2(                                                                       
[...]

PR #654, opened 2024-06-03
Update Gemmini schedules

yamaguchi1024:
Generalize 512x512x512 Gemmini schedules to handle 256 multiples of N and M, which was necessary for producing the [performance results](https://docs.google.com/spreadsheets/d/1oNiDLUAn_DfDARSMfhJ72V3mUMZxT_oNVVYPGH07gtM/edit?usp=sharing)


tests/pldi24/test_gemmini_matmul_ae_rewrite.py
```diff
@@ -102,9 +104,9 @@ def test_matmul_ae(golden):
     res_alloc = res_load.prev()
 
     # Tile loops for a scratchpad and an accumulator
-    gemmini, _ = tile_loops_top_down(gemmini, [(i_loop, 16), (j_loop, 16)])
-    gemmini, [_, j_outer] = tile_loops_top_down(gemmini, [(i_loop, 16), (j_loop, 16)])
-    gemmini, _ = tile_loops_top_down(gemmini, [(k_loop, 16)])
+    gemmini, _ = perfect_tile(gemmini, [(i_loop, 16), (j_loop, 16)])
```
skeqiqevian, line 107: Previously, `tile_loops_top_down` worked for both `conv` and `matmul` Gemmini schedules. Does `perfect_tile` not apply to the `conv` schedule?

skeqiqevian (approved):
Schedule-wise LGTM. As far as I can tell, in addition to generalizing M and N to any multiple of 256, you were also to reduce the accumlator's accumulation size. It wasn't clear to me what schedule changes enabled that optimization, so an explanation would be helpful!

tests/pldi24/gemmini_schedules.py
```diff
@@ -139,10 +139,10 @@ def add_guard(p, c):
         if not isinstance(c, pc.ForCursor):
             break
         try:
-            hi = c.hi().value()
+            hi = expr_to_string(c.hi())
             name = c.name()
             child = p.forward(c).body()[0]
-            p = remove_loop(p, c)
+            p = remove_loop(p, c, unsafe_disable_check=True)
```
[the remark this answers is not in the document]
yamaguchi1024: #652 is the issue about the remove_loop failure

tests/pldi24/test_gemmini_matmul_ae_rewrite.py
yamaguchi1024: `perfect_tile` assumes that the loop is perfectly divisible even when the loop bound is not constant, but this makes me realize that we can just `try/except` that in `tile_loops_top_down` too. Maybe I'll change that

Issue #565, opened 2024-01-31
No-op Intrinsic in Exo 


SamirDroubi:
There has been various examples of situations where we would like to insert no-ops into Exo programs. GPU synchronization #547 is one example. There are some examples in CPUs (prefetching, barriers, pause, etc). Some other examples are architecture independent (inserting debug information into the programs).

I think there are a few questions to answer with regard to this feature:
1. What do we want out of them?
2. How should it be exposed in the IR?
3. How should codegen treat them?
4. How should we be able to insert it into programs?
5. How should other rewrites interact with them?
6. What do no-ops imply from an equivalence/effects perspective?
7. What do no-ops imply from a safety perspective?

I think it is useful to have a discussion on what various things we might want from them and from different perspectives so that we can get a general design that is not overfit to one particular use case. 

-------------------------------------------
**prefetching:**
Here are the answers to the questions above driven by ideas I have had to support prefetching on CPUs and some of the issues that show up:

1. I would like to be able to insert prefetch instructions at random points in the code with a memory operand and other arguments (e.g. locality hints). I would want future rewrites to continuously update the memory operand. I would want the rewrites to ensure that the memory operand always references a buffer that is live. In addition, I want want to be able to perform out-of-bound accesses to the buffer. 
2. My idea is to have a reserved keywords in the compiler e.g. `no-op` that can be used as a name to a proc call. This proc can accept an arbitrary number of arguments.
3. Unreplaced no-op calls are simply no-operations and the arguments are expressions in Exo which have no side-effects. So, we can simply skip them at codegen.
4. There should be a rewrite `insert_no-op(Proc, Gap, *NewExprs)` which inserts a `no-op` call at the gap cursor with the arguments being the provided list of new expressions.
5. Rewrites should continuously update the arguments of this proc call. Just like any other proc call.
6. No-ops should have no effects. What about effects of other parts of the program on the arguments of the no-op? That probably shouldn't matter since the proc itself is a no-op. So, I think effect analysis can generally skip no-op calls.
7. What safety guarantees should be respected? Here are the safety restrictions that I currently know of in Exo:
    - Bounds check: We want to be able to specify out-of-bounds memory operands for prefetching. But before that, what is this memory operand going to be? Is it a read expression or a window expression or something else? It is really neither, but we currently have no other way of specify a location with respect to a buffer.
    - Aliasing: should arguments to no-ops be allowed to alias? I don't have a good example of why this should be relaxed here. So, maybe they shouldn't.
[...]

PR #661, opened 2024-06-07
Implemented `insert_noop_call` scheduling operation

skeqiqevian:
implement a new scheduling op to insert no-op function calls anywhere, addressing #565. This scheduling op will be useful for inserting prefetching and potentially synchronization primitives arbitrarily. 

For now, it only works if the body of the proc is a single `pass` statement, but we could potentially extend this in the future.


src/exo/LoopIR_scheduling.py
```diff
@@ -2738,6 +2739,54 @@ def DoInsertPass(gap):
     return ir, fwd
 
 
+def DoInsertNoopCall(gap, proc, args):
+    srcinfo = gap.parent()._node.srcinfo
+
+    body = proc.body
+    if not (len(body) == 1 and isinstance(body[0], LoopIR.Pass)):
+        # TODO: We should allow for a more general case, e.g. loops of passes
+        raise SchedulingError("Cannot insert a proc whose body is not pass")
```
yamaguchi1024, line 2748: Can you add a test for this error?

yamaguchi1024:
@skeqiqevian Implementation of `insert_noop_call` looks good (I like the typecheck refactoring!) but how does changing the schedule of `apps/x86/sgemm/sgemm.py` affect the performance?

yamaguchi1024 (approved):

PR #672, opened 2024-06-25
Refactor old effects into standalone boundscheck

### LINKED ISSUE #245: Deprecate effects if it's not used, or re-work analysis to use it

yamaguchi1024:
Fixes #245


tests/test_effects.py
skeqiqevian, line 1: We should rename this file to indicate that its just doing bounds checks. I personally in favor of deleting all the golden files because this file isn't testing pretty printing, and it's just clutter in our repository.

tests/golden/test_bounds/test_infereffects.txt
skeqiqevian, line 1: Shouldn't be named infer effects.

tests/test_bounds.py
```diff
@@ -557,7 +515,7 @@ def foo(n: size, m: size, A: f32[n, m]):
     def bar(A: f32[10, 20]):
         foo(A)
 
-    assert bar.show_effects() == golden
+    assert str(bar) == golden
```
skeqiqevian, line 518: This test doesn't test the partial eval at all. Shouldn't we just delete the golden for it?

src/exo/LoopIR_scheduling.py
```diff
@@ -2826,10 +2816,8 @@ def check_pred(pred):
 
         eff = None
         # TODO: raise NotImplementedError("need to figure out effect of new closure")
-        subproc_ir = LoopIR.proc(
-            subproc_name, fnargs, subproc_preds, body, None, eff, info
-        )
-        call = LoopIR.Call(subproc_ir, args, None, info)
+        subproc_ir = LoopIR.proc(subproc_name, fnargs, subproc_preds, body, eff, info)
```
skeqiqevian, line 2819: This should be None, not eff, right?

skeqiqevian: We should add a test for it, I'm surprised it wasn't caught by anything.

skeqiqevian (changes requested):
Requested some minor changes. I did not review `LoopIR_Dependencies` from `LoopIR.py` and most of `boundscheck.py`. For those, I suggest looking at the code coverage – there are some untested new lines of code that might be worth adding tests for. Otherwise, I will trust your judgement on those chunks of code.

skeqiqevian (approved):
Looks good now.

PR #673, opened 2024-06-27
Fix bugs for loop dependent configuration states

yamaguchi1024:
- Add dependency analysis to typecheck for checking loop dependent configuration states, and emit TypeError
- Add type checking to bind_config and write_config


skeqiqevian (changes requested):
Confused about a few things, mostly related to error messages. But also I'm not sure why `size` is not permitted for configuration state. Aren't we only concerned about iteration variables? 

Also, unrelated to this PR, but couldn't `test_write_loop_syntax_check_fail` in `test_config.py` be resolved by first calling simplify?

skeqiqevian (approved):

PR #696, opened 2024-08-23
Improves the error message of invalid index expression use

rtzam:
This commit adds a test to verify that when a
user attempts to allocate a temporary
variable with an index type, a descriptive
error is provided


yamaguchi1024:
@rtzam lgtm, feel free to merge it!

PR #718, opened 2024-10-10
update some code in /apps/gemini/src/platform

xiuhu17:


yamaguchi1024:
Hi @xiuhu17 ! Thanks for the PR. Could you explain why you had to change the Gemmini malloc code? Was it incorrect in some cases? Could you provide a reproducer?

xiuhu17:
Hi @yamaguchi1024 , the reason I change it because the original code looks like: 
```
uint32_t loc = 0;
  if (i > 0) {
    loc = ACC_BLOCKS[i - 1].loc + ACC_BLOCKS[i - 1].size;
  }
  ACC_BLOCKS[i].size = size;
  ACC_BLOCKS[i].loc = 0;
```

The `loc` seems to be useless, and the return value will always be `0x80000000`.

yamaguchi1024:
Sorry for dropping the ball on this! The Gemmini malloc files are a bit stale honestly, and I don't think there's any test case that executes this either. I'm really curious how you found this bug. Were you trying to repurpose this code for some other memory allocation?

xiuhu17:
> Sorry for dropping the ball on this! The Gemmini malloc files are a bit stale honestly, and I don't think there's any test case that executes this either. I'm really curious how you found this bug. Were you trying to repurpose this code for some other memory allocation?

Yeh, I previously was trying to figure out the allocation logic used by EXO, I read the code and found that issue. :)

yamaguchi1024 (approved):

yamaguchi1024:
@xiuhu17 great, thanks for the contribution!

PR #720, opened 2024-10-11
Fix cursor printing errors

yamaguchi1024:


akeley98 (approved):
Looks good to me, I played around with some block and gap cursors for my own weird code (the averaging code) and didn't observe any bad formatting.

PR #726, opened 2024-10-17
Fix fissioning of if/else

akeley98:
Fix incorrect fission of if/else statements. See test_if_fission for repro.

(Old behavior: fissioning the body of an if statement causes the orelse part to be duplicated. Fissioning within the orelse causes an exception)


yamaguchi1024 (approved):
Thanks!

yamaguchi1024:
Feel free to merge it whenever you're ready

PR #728, opened 2024-10-19
Reorganize `src/exo` directory to frontend, core, rewrite, and backend

yamaguchi1024:


yamaguchi1024:
@akeley98 @kehemo ping 👀

kehemo (approved):
looks good

akeley98:
This is a breaking change because "public" modules like exo.memory have been moved (exo.core.memory in this case) which can break sample code and tutorials. (Maybe third-party articles as well?) I otherwise don't have much opinion on how to organize files. If this is acceptable to you, this can be merged.

PR #729, opened 2024-10-21
Update Documentation

yamaguchi1024:
Add documentation for design, imports, system overview, externs, inspection, instructions, memories, and object code. Also add the Cursor example in examples/cursors.


docs/System.md
```diff
@@ -0,0 +1,93 @@
+# System Overview
+
+This document provides an overview of the Exo compilation process, as illustrated in Figure 1 of the PLDI'22 paper.
+
+![System overview](images/system-overview.png)
+
+The Exo compiler consists of a frontend and a backend, with user schedules applied in between. The input to the compiler is a set of Exo source files (`*.py`), and the output is generated C code (`*.c`).
+
```
akeley98, line 8: I feel like this doc would have made most sense if we started out with an overview of the different IRs and how they are used. In particular that LoopIR is the "main" IR used for scheduling (in between the frontend i.e. UAST->LoopIR and the backend i.e. LoopIR->C).

docs/System.md
```diff
@@ -0,0 +1,93 @@
...
+## Core
+
+`src/exo/core` defines IRs used in Exo and other core implementations.
+- `LoopIR.py` is the main file that defines IRs (LoopIR, UAST, PAST), and their visitor functions (LoopIR_Do, LoopIR_Rewrite). 
+- `LoopIR_pprint.py` implements a printing procedure for the IRs defined in `LoopIR.py`.
+- `prelude.py` defines `Sym` and `Srcinfo`.
+
+User-defined features like config, externs, and Memory's parent class implementations are in `configs.py`, `extern.py`, and `memory.py`, respectively.
+
+`internal_cursors` defines cursor movements that are used internally by `LoopIR_scheduling` implementations of scheduling primitives.
```
akeley98, line 38: I feel like a bit more background is needed here for internal_cursors to make sense.

docs/object_code.md
```diff
@@ -0,0 +1,272 @@
...
+### Type and Memory Annotations
+
+In Exo, types and memory spaces are explicitly annotated. The syntax is:
+
+```python
+name: type[size] @ memory
+```
+
+- **`name`**: The variable name.
+- **`type`**: The data type (e.g., `i32`, `f32`).
```
akeley98, line 65: Give a list of valid types (or tell where in the code to find the list of valid types). Their names may not be predictable to C programmers.

docs/object_code.md
```diff
@@ -0,0 +1,272 @@
+# Exo Object Code Syntax
```
akeley98, line 1: As we discussed, should bring up array vs window parameter syntax (`f32[...]` vs `[f32][...]`).

docs/object_code.md
```diff
@@ -0,0 +1,272 @@
...
+
+```python
+data: i32[IC, N] @ DRAM
+```
+
+- **`data`**: The name of the argument.
+- **`i32`**: The data type (32-bit integer).
+- **`[IC, N]`**: A 2D array with dimensions `IC` and `N`.
+- **`@ DRAM`**: Specifies that `data` resides in DRAM memory.
+
```
akeley98, line 83: Currently can't call functions that take arrays (instead of windows) from Exo code.

docs/object_code.md
```diff
@@ -0,0 +1,272 @@
...
+  ```python
+  data[c, j + r]
+  ```
+
+## Limitations
+
+Exo has a few limitations that users should be aware of:
+
+1. **Non-affine indexing**: Exo does not support non-affine indexing. This means that any indexing operation must be a linear combination of loop variables and constants. For example, the following expressions are not allowed:
```
akeley98, line 174: Allows quasi-affine indexing? (Modulo operator)

docs/Imports.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+Enables postponed evaluation of type annotations, allowing you to use forward references in type hints without causing issues during runtime. This is necessary to support Exo's `x : f32` syntax.
+
+
+## 2. Core Exo Module
+
+```python
+from exo import *
+```
```
akeley98, line 32: It may be worth mentioning that the `from foo import bar` (or `*`) import is effectively mandatory to use things in Exo procedures, since Exo object code reserves attribute syntax (`foo.bar`) for config ops.

docs/externs.md
```diff
@@ -0,0 +1,159 @@
...
+            f"expected argument to be a real scalar value, but got type {arg_type}"
+        )
+    return arg_type
+```
+
+- Checks that there is exactly one argument.
+- Ensures the argument is a real scalar type (e.g., `float`, `double`).
+- Returns the type of the argument as the return type of the function.
+
+##### `compile(self, args, prim_type)`
```
akeley98, line 67: Elaborate: this returns a string for passing to `str.format`; also, what are is set of arguments (e.g. `prim_type`) that will get passed to `str.format`? Note in particular that `prim_type` does not use the type names for Exo (e.g. `f32`) ... it could be important for the user to know the naming scheme here if the user is doing something clever and not just pasting the `prim_type` directly into generated C code.

docs/externs.md
```diff
@@ -0,0 +1,159 @@
+# Externs
+
+Externs in Exo provide a mechanism to interface with external functions and libraries directly from your Exo code. By defining custom extern functions, you can extend the capabilities of Exo and leverage existing code written in other languages like C or C++. Externs can be used as expressions in your code, particularly on the right-hand side (RHS) of assignment and reduction statements.
+
+## Defining Externs in User Code
+
+Extern functions are defined by subclassing the `Extern` class provided by Exo. This allows you to specify how the extern function should behave, including type checking, compilation, and any global code it might require.
```
akeley98, line 7: Aliasing restriction should be mentioned.

examples/cursors/cursors.py
```diff
@@ -0,0 +1,142 @@
+from __future__ import annotations
```
akeley98, line 1: Good example code!

docs/memories.md
```diff
@@ -0,0 +1,162 @@
...
+- **`can_read(cls)`**: Controls whether the memory can be read directly. Setting it to `False` means you cannot read/write directly to this memory using standard array access.
+
+  ```python
+  @classmethod
+  def can_read(cls):
+      return False
+  ```
+
+- **`alloc(cls, new_name, prim_type, shape, srcinfo)`**: Defines how memory allocation is handled. For `AVX512`, it ensures that the allocated memory represents 16-wide vectors of `float` type.
+
```
akeley98, line 76: Possible values of `prim_type` should be documented.

docs/memories.md
```diff
@@ -0,0 +1,162 @@
...
+  ```python
+  @classmethod
+  def free(cls, new_name, prim_type, shape, srcinfo):
+      return ""
+  ```
+
+- **`window(cls, basetyp, baseptr, indices, strides, srcinfo)`**: Defines how to access elements in the memory.
+
+  ```python
```
akeley98, line 93: The window member functions needs to be explained more thoroughly.

docs/instructions.md
```diff
@@ -0,0 +1,164 @@
+# External Instruction Definitions
```
akeley98, line 1: This file is in good shape, except like before I think `prim_type` should be documented more thoroughly.

examples/cursors/cursors.py
yamaguchi1024: Thanks!

docs/Design.md
```diff
@@ -0,0 +1,69 @@
...
+These user-defined scheduling operations can encapsulate common optimization patterns and hardware-specific transformations, greatly improving productivity. They can be put together in reusable libraries, further enabling modularity and portability.
+
+More infomation can be found in the [ASPLOS paper](.) and [Cursor.md](./Cursor.md).
+
+## The AIR Framework: Action, Inspection, Reference
+
+We identified that Action, Inspection, and Reference are the key scheduling language design mechanisms that enable user-defined scheduling operations.
+
+- **Actions** are the scheduling primitives that transform the code (e.g., `divide_loop`, `reorder`).
+- **Inspections** query properties of the code (e.g., loop bounds, memory access patterns).
```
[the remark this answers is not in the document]
yamaguchi1024: Good point! I added inspection documentation separately.

docs/object_code.md
yamaguchi1024: I don't think this is true?? You should be able to write sth like:
```
@proc
def foo(x : f32[10):
  ...
@proc
def bar(y : f32[10):
  ...
  foo(y)
```

docs/instructions.md
```diff
@@ -0,0 +1,164 @@
...
+def neon_vld_4xf32(dst: [f32][4] @ Neon, src: [f32][4] @ DRAM):
+    assert stride(src, 0) == 1
+    assert stride(dst, 0) == 1
+
+    for i in seq(0, 4):
+        dst[i] = src[i]
+```
+
+- **`@instr("{dst_data} = vld1q_f32(&{src_data});")`**: Specifies the C code to emit when this instruction is called.
+  - `{dst_data}` and `{src_data}` are format strings that will be replaced with the actual arguments during codegen.
```
akeley98, line 46: I think we should leave this for later since we may need to do a bunch of work to fix the issues with window expressions and custom memory types that I identified. But eventually I think we should document the placeholders for `@instr` more.

docs/inspection.md
```diff
@@ -0,0 +1,50 @@
+# External Inspection Functions
```
akeley98, line 1: Good new documentation, thanks.

docs/object_code.md
yamaguchi1024: Ok I'll add this example as well

docs/memories.md
yamaguchi1024: Done!

docs/memories.md
```diff
@@ -0,0 +1,200 @@
...
+    - `f16` -> `"_Float16"`
+    - `f32` -> `"float"`
+    - `f64` -> `"double"`
+    - `i8`  -> `"int8_t"`
+    - `ui8` -> `"uint8_t"`
+    - `ui16`-> `"uint16_t"`
+    - `i32` -> `"int32_t"`
+  - `shape`: A list of C strings representing the shape of each dimension. In the example above, it would be `["N", "M"]`.
+
+  For `AVX512` memory, the `alloc` method ensures that the allocated memory represents 16-wide vectors of the `float` type.
```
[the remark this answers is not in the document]
yamaguchi1024: fixed

docs/instructions.md
yamaguchi1024: Yeah, just telling ppl to use `_data` is probably appropriate here

docs/externs.md
```diff
@@ -0,0 +1,170 @@
...
+##### `__init__(self)`
+
+Initialize your extern function with its name.
+
+```python
+def __init__(self):
+    super().__init__("sin")
+```
+
+- `"sin"`: The name of the external function as it will appear in the generated code.
```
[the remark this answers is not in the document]
yamaguchi1024: Yeah good point

docs/externs.md
```diff
@@ -0,0 +1,170 @@
...
+  - `i32` -> `"int32_t"`
+
+```python
+def compile(self, args, prim_type):
+    return f"sin(({prim_type}){args[0]})"
+```
+
+- Generates the code that calls the external function, ensuring proper casting to the primitive type.
+
+##### `globl(self, prim_type)`
```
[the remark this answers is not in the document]
yamaguchi1024: Good catch! Yeah this is a bug, I filed it here: #737 737

docs/externs.md
```diff
@@ -0,0 +1,170 @@
...
+sin = _Sin()
+```
+
+- `sin` now represents the extern function and can be used like any other expression in Exo.
+
+## Using Externs as Expressions
+
+Externs can be used as expressions on the RHS of assignment and reduction statements. This allows you to incorporate external functions seamlessly into your Exo computations.
+
+Note that externs (and Exo procedures) do not allow aliasing in their arguments. This restriction is in place to prevent externs from having side effects on the input arguments.
```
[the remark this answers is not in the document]
yamaguchi1024: Also, externs actually allow aliasing.
```
@proc
def foo(xo : f32, yo : f64, xi : f32, yi : f64):
    xo = select(xi, xi, xi, xi) 
```

gets compiled into
```
 float _select_float(float x,float v,float y,float z) {
    if (x < v) return y;
    else return z;
}

// foo(
//     xo : f32 @DRAM,
//     yo : f64 @DRAM,
//     xi : f32 @DRAM,
//     yi : f64 @DRAM
// )
void foo( void *ctxt, float* xo, const double* yo, const float* xi, const double* yi ) { 
*xo = _select_float((float)*xi, (float)*xi, (float)*xi, (float)*xi);
}
```

akeley98 (approved):

PR #761, opened 2025-02-21
Fix various small bugs in new_eff that were failing to handle window

### LINKED ISSUE #594: WindowStmt(s) aren't treated as declarative statements within the compiler
Must of our rewrites pay attention to liveness issue of variables declared by alloc statements. However, we don't ensure similar checks for WindowStmts. This means that a schedule like the following:

```
def test_fission_after_window_stmt(golden):
    @proc
    def foo(n: size, res: f32, x: f32[1]):
        for i in seq(0, n):
            win = x[0:1]
            res += win[0]

    foo = fission(foo, foo.find("win = _").after())
```
Can result in the following:
```
def foo(n: size, res: f32 @ DRAM, x: f32[1] @ DRAM):
    for i in seq(0, n):
        win = x[0:1]
    for i in seq(0, n):
        res += win[0]
```
which is incorrect because `win` is not live at the read.

There are multiple solutions to this:
- Go and make sure that we also consider WindowStmt anywhere we do something special for an Alloc.
- Expand the semantics of Alloc to be a decelerative statements for allocations and Windows: `win: [f32][1]; win = x[0:1]`
- Reassess how important it is to have a WindowStmt. My impression is that this is not something that you naturally use, it often results from inlining procedure which we often follow by inlining the windows manually, but you could also imagine inlining the windows as well by default. I guess I don't think I have seen an example where the existence of a WindowStmt has been valuable and I wonder if it makes sense to remove it for the sake of simplifying the IR and reducing the number of statements you have to worry about when writing user-code.

yamaguchi1024:
Fixes #594

### DIFF
--- a/src/exo/rewrite/new_eff.py
+++ b/src/exo/rewrite/new_eff.py
@@ -1100,27 +1100,6 @@ def get_changing_globset(env):
 # Extraction of Effects from programs
 
 
-def window_effs(e):
-    eff_access = []
-    syms = {}
-    for i, w in enumerate(e.idx):
-        if isinstance(w, LoopIR.Interval):
-            syms[i] = Sym(f"EXO_EFFECTS_WINDOW_TEMP_INDEX_{i}")
-            eff_access.append(lift_e(syms[i]))
-        else:
-            eff_access.append(lift_e(w.pt))
-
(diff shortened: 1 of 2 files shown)


yamaguchi1024:
stage_mem is not fixed in this PR, should probably fix it, too

PR #769, opened 2025-03-14
Fix typos.

jasondavies:


yamaguchi1024:
@jasondavies Thanks for catching the typos!

PR #777, opened 2025-04-21
Add Arm SVE instructions support

eyr1n:
# Overview

Add support for Arm SVE instructions to the Exo language.

# Arm SVE Instructions

A SIMD instructions with registers of variable bit widths depending on the implementation. 

Two vectorization approaches are available:

- Vector Length Specific (VLS, traditional SIMD)
- Vector Length Agnostic (VLA, can be applied to variable length vectors)

# platforms/sve_vls.py

This file implements the VLS approach. 

`Memory` class and `@instr` function require knowledge of the target processor's register bit width before the Exo parser operates. By wrapping them in `SVE_VLS` class, their initialization and the operation of the parser is deferred.

Example:

```py
SVE_VLS = SVE_VLS(512)

# ...

vls = set_memory(vls, "SVE_reg:_", SVE_VLS.Vector)
vls = replace_all(vls, SVE_VLS.svld1_f32)
```

# platforms/sve_vla.py

This file implements the VLA approach.

`@instr` functions don't require knowledge of the target processor's register bit width.

Example:

```py
vla = replace_all(vla, svmla_n_f32_x_vla)
```

To use the output generated by Exo compilation, `libs/exo_arm_sve.h` is required.

### DIFF
--- a/src/exo/libs/exo_arm_sve.h
+++ b/src/exo/libs/exo_arm_sve.h
@@ -1,0 +1,28 @@
+#ifndef EXO_ARM_SVE_H
+#define EXO_ARM_SVE_H
+
+#include <arm_sve.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static inline __attribute__((always_inline)) void
+svmla_n_f32_x_vla(int64_t N, float32_t *dst, const float32_t *src1,
+                  float32_t src2) {
+  int64_t i = 0;
+  svbool_t pg = svwhilelt_b32(i, N);
+  do {
+    svst1_f32(pg, &dst[i],
+              svmla_n_f32_x(pg, svld1_f32(pg, &dst[i]), svld1_f32(pg, &src1[i]),
+                            src2));
+    i += svcntw();
+    pg = svwhilelt_b32(i, N);
+  } while (svptest_first(svptrue_b32(), pg));
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/src/exo/platforms/sve_vls.py
+++ b/src/exo/platforms/sve_vls.py
@@ -22,11 +22,8 @@ def _sve_vector_init(sve_vector_bits):
         @classmethod
         def global_(cls):
             return f"""#include <arm_sve.h>
-#ifdef __FUJITSU
-typedef svfloat32_t svfloat32_vls_t;
-#else
 typedef svfloat32_t svfloat32_vls_t __attribute__((arm_sve_vector_bits({sve_vector_bits})));
-#endif"""
+typedef svfloat64_t svfloat64_vls_t __attribute__((arm_sve_vector_bits({sve_vector_bits})));"""
 
         @classmethod
         def alloc(cls, new_name, prim_type, shape, srcinfo):
--- a/src/exo/platforms/sve_vls.py
+++ b/src/exo/platforms/sve_vls.py
@@ -78,15 +78,14 @@ class SVE_VLS:
 
         @instr("{dst_data} = svld1_f32(svptrue_b32(), &{src_data});")
         def svld1_f32(
-            N: size,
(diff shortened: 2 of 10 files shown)


src/exo/libs/exo_arm_sve.h
```diff
@@ -0,0 +1,28 @@
...
+#define EXO_ARM_SVE_H
+
+#include <arm_sve.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static inline __attribute__((always_inline)) void
+svmla_n_f32_x_vla(int64_t N, float32_t *dst, const float32_t *src1,
```
yamaguchi1024, line 11: Did you have to define this wrapper instruction for VLA because it wasn't really possible to program the instruction directly in Exo?

src/exo/platforms/sve_vls.py
```diff
@@ -0,0 +1,119 @@
...
+    vec_types = {
+        "float": (sve_vector_bits // 32, "svfloat32_vls_t"),
+        "double": (sve_vector_bits // 64, "svfloat64_vls_t"),
+    }
+
+    class SVE_VLS(Memory):
+        @classmethod
+        def global_(cls):
+            return f"""#include <arm_sve.h>
+#ifdef __FUJITSU
```
yamaguchi1024, line 25: This makes me wonder, how much of this PR FUJITSU specific? Is this going to work on non-FUJITSU SVE machines?

yamaguchi1024:
@eyr1n Thanks for the PR! I had some questions and suggestions but it basically looks ok

yamaguchi1024:
There seems to be a linting error. Did you `pre-commit install` before your commit and apply code formatting?

src/exo/platforms/sve_vls.py
eyr1n: Sorry, this line is my mistake.
`__FUJITSU` is defined by Fujitsu C Compiler, but Fujitsu C Compiler cannot handle VLS.
I'll remove it.

eyr1n:
> There seems to be a linting error. Did you `pre-commit install` before your commit and apply code formatting?

I forgot to prepare pre-commit, but the error still appears.
linting seems to have failed in the checkout process.

src/exo/libs/exo_arm_sve.h
eyr1n: VLA should be usable even when the SVE register width of the target processor is unknown. Actually, Arm provides instructions to get the register width dynamically.
When using the instructions with SVE predicate registers to process an array, I think this loop cannot be fully expressed in Exo (as far as I know).

yamaguchi1024: We're working on supporting the vector length agnostic RVV instruction directly with Exo, and it'll be interesting to support SVE as well at some point. But that's for future work

src/exo/platforms/sve_vls.py
```diff
@@ -0,0 +1,116 @@
...
+    vec_types = {
+        "float": (sve_vector_bits // 32, "svfloat32_vls_t"),
+        "double": (sve_vector_bits // 64, "svfloat64_vls_t"),
+    }
+
+    class SVE_VLS(Memory):
+        @classmethod
+        def global_(cls):
+            return f"""#include <arm_sve.h>
+typedef svfloat32_t svfloat32_vls_t __attribute__((arm_sve_vector_bits({sve_vector_bits})));
```
yamaguchi1024, line 25: The reason you're getting a Codecov warning is that you don't have tests that compile the scheduled Exo code.
Can you use `compiler` (refer to test_codegen.py) to generate C code and add that to golden as well? We want to test the generated C code as well as Exo IR.

yamaguchi1024: @eyr1n lmk if you have time to make this change! The PR looks good otherwise.

PR #800, opened 2025-07-25
A little loving maintenance help

alexreinking:
This PR cleans up some of the annoying software process stuff I'm sure none of you want to deal with 🙂 

1. I moved as much as possible into the pyproject.toml, so no more `setup.py`, `setup.cfg`, `MANIFEST.in`, or `tox.ini`.
2. I cleaned up the specification of dependencies within the pyproject.toml file to name a consistent set.
    1. The dependencies in `[project]` in the `dependencies` group are for downstreams, too.
    2. The dependencies in `[dependency-groups]` under `dev` are required only to develop exo.
    3. The dependencies in `[dependency-groups]` under `tools` are command-line tools that are likely to conflict with system commands; these are not installed by default when using `uv`, but are included in the `requirements.txt` to maintain compatibility with existing workflows.
4. I replaced the `requirements.txt` file with a "lock file" version -- this is automatically generated and ensures a consistent and compatible set of dependencies across Python versions. Think of it like so:
    1. The logical constraints belong in `pyproject.toml`
    2. A particular solution to those constraints at some moment in time belongs in `requirements.txt`
5. One test was broken on newer Python versions because the text in `NameError` changed. I patched that test.
6. Some tests threw warnings. I cleaned up those warnings.
7. I upgraded Black to the latest version and brought the whole codebase up to date. This mostly affected redundant parens and blank lines, but in a small number of cases, I restructured the code to be both more readable and amenable to Black's formatting. I also disabled Black in one block of code where the formatting was just so.
8. The minimum CMake version was inconsistent across the project between `3.21` and `3.22` -- I standardized everything on `3.22` to be minimally disruptive.

---

To make changes to the `requirements.txt` file, start by modifying `pyproject.toml` with your desired new packages or version constraints. Then, using `uv`, run:

```
$ uv export --all-groups --no-hashes --no-emit-project -o requirements.txt
```

You can install the new requirements via:

```
$ pip install -r requirements.txt
```

Prefix that command with `uv` to make it much, much faster.


alexreinking:
I just cleaned up the history so that each commit makes sense on its own (albeit in sequence). That will hopefully make this easier to review.

alexreinking:
Do you have any questions or concerns about this PR?

PR #807, opened 2025-10-14
Widen the column limit in yapf

alexreinking:
There was some discussion in the slack that the automatically formatted Exo code is hard to read. Here's a PR that at least widens the column limit, but also consolidates the formatting operation down to a single private function so you can do whatever you want there.

As a reminder, updating the goldens is done by:

```console
$ pytest tests --update-golden -n auto
```

Hope this greases the wheels.


yamaguchi1024 (approved):
LGTM, thanks Alex!

PR #812, opened 2025-11-17
Fix a bug that caused false type errors for externs

jiriklepl:
Hi,

I've used Exo in some experiments and encountered an issue with precision analysis when using `LoopIR.Extern`.

The issue happens at src/exo/backend/prec_analysis.py:202-222 when none of the arguments in `e.args` have a resolved type without passing them through `self.apply_e`. This leads to `typ == T.R` at the beginning of the next loop (starting at 209), which breaks its logic.

This pull request seems to be the most conservative resolution of the issue - it simply reorders passing the arguments through `self.apply_e` before computing the shared `typ`.

The change doesn't break any tests in `pytest`.

### DIFF
--- a/src/exo/backend/prec_analysis.py
+++ b/src/exo/backend/prec_analysis.py
@@ -200,24 +200,23 @@ class PrecisionAnalysis(LoopIR_Rewrite):
             return LoopIR.BinOp(e.op, lhs, rhs, typ, e.srcinfo)
 
         elif isinstance(e, LoopIR.Extern):
+            new_args = [self.apply_e(a) for a in e.args]
+
             typ = T.R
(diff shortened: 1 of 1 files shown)


yamaguchi1024 (approved):
@jiriklepl Thanks for the PR and sorry for the late response! The change LGTM.

yamaguchi1024:
Linting is failing, likely due to an external contribution, but I’ll merge this once the other CI checks pass

yamaguchi1024:
@jiriklepl merged, thanks again for the contribution!

jiriklepl:
@yamaguchi1024 glad you like the changes, and thank you for merging them in!

PR #835, opened 2026-06-16
Fix frame.f_locals usage for PEP 667 (Python 3.13+)

sueszli:
[PEP 667](https://peps.python.org/pep-0667/) (Python 3.13+): `frame.f_locals` returns a `FrameLocalsProxy`, not a `dict`. Snapshot via `dict(...)` at the three scope-capture sites.

Adapted from https://github.com/sueszli/exojit/blob/main/exojit/patches_exo.py.

(Happy to migrate the 3 other patches as well!)

PR #836, opened 2026-06-16
Fix lint workflow checkout for fork PRs

sueszli:
`Lint` runs on `pull_request_target` but `actions/checkout@v4` only sets `ref:`, so it defaults to `exo-lang/exo` and can't find the fork's branch. Every external fork PR fails (see #835).

Set `repository: head.repo.full_name` and pin to `head.sha`.

Safe: the lint steps only read files, no repo code runs.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -7,7 +7,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


yamaguchi1024 (approved):
This is great, thank you @sueszli !

sueszli:
Thanks for reviewing!

